PR #47, opened 2015-06-16
Fixes from OSPC meeting

evan-magnusson:


kerkphil:
I think NOT.  Actually, change that.  I'm OK with this change.

Python/TESTED--SSmethod2/SS.py
```diff
@@ -177,7 +177,7 @@ def new_SS_Solver(b_guess_init, n_guess_init, wguess, rguess, T_Hguess, factorgu
         factor = misc_funcs.convex_combo(new_factor, factor, params)
         T_H = misc_funcs.convex_combo(new_T_H, T_H, params)
         
-        dist = np.array([misc_funcs.perc_dif_func(new_r, r)] + [misc_funcs.perc_dif_func(new_w, w)] + [misc_funcs.perc_dif_func(new_T_H, T_H)]).max()
+        dist = np.array([misc_funcs.perc_dif_func(new_r, r)] + [misc_funcs.perc_dif_func(new_w, w)] + [misc_funcs.perc_dif_func(new_T_H, T_H)] + [misc_funcs.perc_dif_func(new_factor, factor)]).max()
```
rickecon: @evan-magnusson, what does that addition do?

Python/TESTED--SSmethod2/run_simulation.py
```diff
@@ -27,7 +27,6 @@
 import cPickle as pickle
 import os
 from glob import glob
-import sys
```
rickecon: @evan-magnusson , Are we not using "import sys" in ths file, or did we want to change something in the body of the file?

rickecon:
Looks good.

Python/TESTED--SSmethod2/SS.py
evan-magnusson: We want to make sure that the percent difference in the scaling factor also converges.

PR #59, opened 2015-06-19
Documentation

evan-magnusson:
Documented some functions in the python files, updated what files are generated/used by each python file.  Also, where graphing occurs conditionally, I put the import statements for matplotlib within those functions, so we don't import them if we don't want them.


isaacswift:
Looks good.  Can't have too much documentation.

PR #64, opened 2015-06-22
Step 1 of Jason's doc

isaacswift:
I did some debugging of the code.  The code is running, but it doesn't converge.  @jdebacker  let me know if this looks like the right idea.  @evan-magnusson if you could help with debugging.

### DIFF
--- a/Python/DEVELOPMENT/SS.py
+++ b/Python/DEVELOPMENT/SS.py
@@ -138,10 +138,9 @@ def Euler_equation_solver(guesses, r, w, T_H, factor, j, params, chi_b, chi_n, t
     error1[mask4] += 1e14
     return list(error1.flatten()) + list(error2.flatten())
 
-def wrguess(X, bssmat, nssmat, j, params, chi_b, chi_n, tau_bq, rho, lambdas, weights, e):
+def wrguess(X, bssmat, nssmat, params, chi_b, chi_n, tau_bq, rho, lambdas, weights, e):
     
     J, S, T, beta, sigma, alpha, Z, delta, ltilde, nu, g_y, tau_payroll, retire, mean_income_data, a_tax_income, b_tax_income, c_tax_income, d_tax_income, h_wealth, p_wealth, m_wealth, b_ellipse, upsilon = params
-
     w, r, T_H, factor = X
 
     for j in xrange(J):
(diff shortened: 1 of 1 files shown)


jdebacker:
This is what I was thinking.  I'm not seeing anything that would suggest it wouldn't converge.  Should we try saving the errors and guesses (r, w, TH, factor) and plotting?  Maybe we can trace out the excess demand functions and find some information there (e.g. local min or very flat functions)?

jdebacker:
One idea - the model hasn't changed, just the solution.   What do the
excess demands look like with the "old" equilibrium values of r, w, TH,
factor?

jdebacker:
Another idea - set delta = 0 and see how it works. This is the part I'm
most unsure about.

evan-magnusson:
Setting delta = 0 doesn't change anything too much.  It still doesn't converge.

jdebacker:
Hmm, not sure.  With delta = 0, then C=Y (the capital stock is unchanged each period).  And we should be able to use HH demand for C (which we can find given r,w,TH,factor) to find demand for K and L from the firm FOCs (using the guesses at r and w).  

It would seem there is a fixed point in here where the factor demands equal factor supplies.  I can imagine that it's hard to find, but I can't see what it wouldn't exist.  

Back to the plotting suggestion - it can be useful just to see if the excess demand are monotonic in r, w, TH, factor.  I'd guess that excess demand for capital is monotonic in r and excess demand for labor is monotonic in w, but it's less clear how the other eq'm variables affect these.  It's also not clear how w affects excess demand in the capital market.  All this is to say that the eq'm might be tough to find.

Fullerton and Rogers (1993) describe finding the eq'm in this way.  Their book not very recent, but they mention this paper https://ecompapers.biz.uwa.edu.au/paper/PDF%20of%20Discussion%20Papers/1982/82-05.pdf  However, after scanning the paper, I didn't see anything especially relevant for this issue.  They also mention the "Scarf's grid search algorithm" and and dissertation chapter by an O. Merrill that I couldn't find as two alternative algorithms.  I don't know anything about these.

jdebacker:
Anything weird going on with the stationarization?  Do we need to change
everything to move from using the firm FOCs from getting prices to getting
quantities?

PR #68, opened 2015-06-23
Firmstep1

evan-magnusson:


jdebacker:
How's this working?  Are we getting an eq'm?

evan-magnusson:
Still not converging.

rickecon:
This seems like a theory issue. We should look at the theory to make sure everything is all closed up. I should be able to get to that by this evening.

PR #71, opened 2015-06-24
TPI debug

evan-magnusson:
TPI has had 4 issues in the first few periods that we couldn't figure out.  They were:

-strange jump after first period in K and L timepaths
-the first upper triangle of the twist doughnut wouldn't work with scalar values
-consumption was negative in a couple entries
-consumption had discontinuous jumps for a couple individuals

These have all been fixed.  The first was due to grabbing the wrong capital stock for the upper triangle of TPI.  The second just needed us to finish writing a scalar version of the total_tax function.  The final was due to some poor initial guesses in TPI.


evan-magnusson:
Also, this does not affect our wealth tax results at all.  Only the first few time periods of TPI were affected.  But this change has been updated in the SS_method2 folder of the wealth tax repo.

PR #72, opened 2015-06-24
uncommented graph file

evan-magnusson:


talumbau:
Is the idea for this to be covered by a flag now?

evan-magnusson:
For TPI_graphs.py and SS_graphs.py, we certainly could include them in the run_baseline.py file and include flags if we want them to print baseline or baseline and experiment graphs.  However, for now we have only run them separately from everything else.  I just forgot to uncomment out a few lines (that referenced pickles that did not exist because I didn't run an experiment).

PR #74, opened 2015-06-25
TPI array indexing fix

isaacswift:
Made nicer array indexing the way that I discussed with @evan-magnusson .


talumbau:
It's definitely more efficient to pull the declaration of `b_mat` and `n_mat` up to the top, and it looks like the value of either is not used until lines 323 and 325. The difference is that, in the current version, the values that are read in lines 323 and 325 come from a `b_mat` and `n_mat` that came into existence only at the last iteration of the while loop, and in the proposed change, the `b_mat` and `n_mat` have been overwritten for every iteration of the while loop. It seems to always write to the same indices though, so it looks fine to me. Just wanted to point it out though to make sure you are OK with that.

isaacswift:
I'm aware.

evan-magnusson:
I think there is an error -- do you need to redefine your ind2 within the loop after line 367?

isaacswift:
Yep.  Got it.

PR #81, opened 2015-07-01
Firmprob

evan-magnusson:
Many changes to the firmproblem folder, fixing numerous bugs with the bequests, and changing the structure slightly.

### DIFF
--- a/Python/Development_firmproblem/SS.py
+++ b/Python/Development_firmproblem/SS.py
@@ -11,8 +11,6 @@ This py-file calls the following other file(s):
             misc_funcs.py
             OUTPUT/Saved_moments/params_given.pkl
             OUTPUT/Saved_moments/params_changed.pkl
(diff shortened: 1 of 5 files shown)


rickecon:
We are still working on this. It is not done and does not work yet. But the changes are something that I want to merge into my fork and work on with @evan-magnusson . Once this works, we will submit a pull request and then make the changes to the package that @talumbau put together.

rickecon:
@evan-magnusson and I just finished going line-by-line through the "Development_firmproblem" folder. We added 1/(1+g_n) to the BQ equation and the K equation. We found one little error where an index was off by one period. But we still have the resource constraint issue. @evan-magnusson and I will spend this afternoon going through an early version of the code for a simpler model--with endog labor, bequest motive, pop growth, no taxes--that has the same resource constraint issue. @jdebacker

PR #88, opened 2015-07-14
[Development_firmproblem] [WIP] Multitude of bug fixes, mostly involving resource constraint violations

evan-magnusson:
Here is a list of everything that changed:
- [x] Add g_n as an input to the get_K and get_BQ functions
- [x] put g_n_ss into parameter vector (and update all functions that unpack this) so that it isn't a global
- [x] put g_n vector into TPI calls of get_K and get_BQ
- [x] make sure K and BQ are done correctly in graphing files
- [x] separate omega from lambda (this will involve adding inputs to a few functions)
- [x] use the correct stationarized resource constraint in SS, TPI, SSgraphs and TPIgraphs
- [x] make aggC and aggI functions
- [x] shift the immigration rates: make the last one 0
- [x] zeroed out immigration rates in demographics.py
- [x] make the last survival rate 0 inside of demographics
- [x] output rho from demographics, don't compute it in run_simulation
- [x] test SS and TPI to make sure resource constraint is satisfied
- [x] test both graphing files to make sure all new broadcasting and shapes work


Python/Development_firmproblem/SS.py
```diff
@@ -102,22 +101,22 @@
 income_tax_params = [a_tax_income, b_tax_income, c_tax_income, d_tax_income]
 wealth_tax_params = [h_wealth, p_wealth, m_wealth]
 ellipse_params = [b_ellipse, upsilon]
-parameters = [J, S, T, beta, sigma, alpha, Z, delta, ltilde, nu, g_y, tau_payroll, retire, mean_income_data] + income_tax_params + wealth_tax_params + ellipse_params
+parameters = [J, S, T, beta, sigma, alpha, Z, delta, ltilde, nu, g_y, g_n_ss, tau_payroll, retire, mean_income_data] + income_tax_params + wealth_tax_params + ellipse_params
```
evan-magnusson: Because it was never used, we didn't keep track of this parameter.  Now we need to.

evan-magnusson:
You'll see a lot of reshaping with the omega's and lambdas.  Some of this is probably unneeded, particularly with the lambdas.  I'm still getting used to knowing when Python will know how to broadcast correctly and when it won't.

Python/Development_firmproblem/SS_graphs.py
```diff
@@ -26,6 +26,9 @@
 from mpl_toolkits.mplot3d import Axes3D
 import cPickle as pickle
 
+import firm_funcs as firm
+import household_funcs as house
```
evan-magnusson: We need to import these functions, as the graphing files now generate the Iss and Css variables.  We could just pickle these, and not need to import anything.  Either way.

Python/Development_firmproblem/TPI.py
```diff
@@ -408,12 +408,17 @@ def Steady_state_TPI_solver(guesses, winit, rinit, BQinit, T_H_init, factor, j,
 
 tax_path = tax.total_taxes(rinit[:T].reshape(T, 1, 1), b_s, winit[:T].reshape(T, 1, 1), e.reshape(
     1, S, J), n_mat[:T], BQinit[:T, :].reshape(T, 1, J), lambdas, factor_ss, T_H_init[:T].reshape(T, 1, 1), None, 'TPI', False, parameters, theta, tau_bq)
-cinit = house.get_cons(rinit[:T].reshape(T, 1, 1), b_s, winit[:T].reshape(T, 1, 1), e.reshape(1, S, J), n_mat[:T], BQinit[:T].reshape(T, 1, J), lambdas.reshape(1, 1, J), b_splus1, parameters, tax_path)
+c_path = house.get_cons(rinit[:T].reshape(T, 1, 1), b_s, winit[:T].reshape(T, 1, 1), e.reshape(1, S, J), n_mat[:T], BQinit[:T].reshape(T, 1, J), lambdas.reshape(1, 1, J), b_splus1, parameters, tax_path)
```
evan-magnusson: rename cinit to a less confusing name, c_path

Python/Development_firmproblem/demographics.py
```diff
@@ -422,12 +416,12 @@ def get_omega(S, J, T, bin_weights, starting_age, ending_age, E, flag_graphs):
         starting_age, ending_age, S+1), poly_int_pop)
     new_omega = pop_int[1:]-pop_int[:-1]
     surv_array, children_rate = get_survival(S, starting_age, ending_age, E)
+    surv_array[-1] = 0.0
     imm_array, children_im = get_immigration2(S, starting_age, ending_age, E)
+    imm_array *= 0.0
```
evan-magnusson: Here is the bandaid.  Eventually, this line will disappear, once we find the immigration rates bug.

evan-magnusson:
Let's not merge it quite yet, I still need to test TPI and make sure the resource constraint is fine in all periods.

evan-magnusson:
I ran a full TPI, and looked out how the resource constraint holds across the T periods.  The difference was around 1 x 10^-6 (or smaller in some instances).  This worried me at first, but then I realized that the minimum distance for the while loop in TPI is around 10^-6, much larger than the 10^-9 used in the SS while loop (which has a much smaller difference in the resource constraint).  If we use a larger minimum distance in the SS while loop, it also has a larger difference in the resource constraint.  I looked through the TPI code, and everything looks good, so I'm assuming the difference just has to do with the high mindist.

Issue #85, opened 2015-07-10
Merge all Python folders into one


evan-magnusson:
Instead of having tested and multiple development folders for python code, we'll want to just have one folder where the code is stored, and utilize branches to handle different versions/projects. 
Checklist:
- [x] Get Development_firmproblem up to date, with all bug fixes finished and solved
- [x] Delete the Development_fsolve and TESTED_DONT_TOUCH folders
- [x] Transfer the changes in Development_firmproblem into the dynamic folder that @talumbau is working on, since that code will not have many of the changes we've implemented in the past few weeks
- [x] Delete CURRENT_DEBUGGED_VERSION folder

evan-magnusson:
@talumbau I'm about finished with the bug fixes in the code -- are you still working on what is in the dynamic folder?  Whenever you're ready, I can get started on moving over the bug fixes.

evan-magnusson:
Being done in PR #94 .

talumbau:
go ahead and have it. I have no outstanding changes to merge in. I can help with review when you are ready for someone to have a look at it. this is good timing as I hope to dig back in to this with additional changes after PyData conference this weekend.

evan-magnusson:
Finished, see PR #94 .

PR #94, opened 2015-07-20
Merging CURRENT_VERSION and DYNAMIC folders

evan-magnusson:
I'm in the process of moving all the changes in the CURRENT_DEBUGGED_VERSION folder into the dynamic folder, as in Issue #85


evan-magnusson:
Items in CURRENT_DEBUGGED_VERSION to be merged over to dynamic:
- [x] income_nopoly.py
- [ ] demographics.py
- [ ] firm_funcs.py
- [ ] household_funcs.py
- [ ] income_polynomials.py
- [ ] labor_data.py
- [ ] misc_funcs.py
- [ ] run_simulation.py
- [ ] SS.py
- [ ] SS_graphs.py
- [ ] tax_funcs.py
- [ ] TPI.py
- [ ] TPI_graphs.py
- [ ] wealth_data.py
- [ ] Make sure the GIF graphing files are stored somewhere
Items in dynamic to make sure are up to date and functioning:
- [x] income_nopoly.py
- [ ] run_simulation.py
- [ ] run_small.py
- [ ] setup.py
- [ ] __init__.py
- [ ] demographics.py
- [ ] firm.py
- [ ] household.py
- [ ] income.py
- [ ] labor.py
- [ ] parameters.py
- [ ] SS.py
- [ ] SS_graphs.py
- [ ] tax.py
- [ ] TPI.py
- [ ] TPI_graphs.py
- [ ] utils.py
- [ ] wealth.py
- [ ] test_basic.py

Python/dynamic/household.py
```diff
@@ -19,38 +21,71 @@
...
-    '''
-    J, S, T, beta, sigma, alpha, Z, delta, ltilde, nu, g_y, tau_payroll, retire, mean_income_data, a_tax_income, b_tax_income, c_tax_income, d_tax_income, h_wealth, p_wealth, m_wealth, b_ellipse, upsilon = params
+    J, S, T, beta, sigma, alpha, Z, delta, ltilde, nu, g_y, g_n_ss, tau_payroll, retire, mean_income_data, a_tax_income, b_tax_income, c_tax_income, d_tax_income, h_wealth, p_wealth, m_wealth, b_ellipse, upsilon = params
     try:
-        deriv = b_ellipse * (1/ltilde) * ((1 - (n / ltilde) ** upsilon) ** (
-            (1/upsilon)-1)) * (n / ltilde) ** (upsilon - 1)
+        deriv = b_ellipse * (1.0/ltilde) * ((1.0 - (n / ltilde) ** upsilon) ** (
+            (1.0/upsilon)-1.0)) * (n / ltilde) ** (upsilon - 1.0)
     except ValueError as ve:
+        # I think TJ added this in.  We need to be careful with doing stuff like this --
```
evan-magnusson: @talumbau @rickecon I'm assuming this addition (the try/except) was added to the code in this function in order to prevent the simulation from failing due to fsolve trying negative values of n.  If we do this, we need to make sure the value that we're setting deriv equal to (i'm assuming this is to prevent it from being a NaN) is not one that is close to a value that is close to a possible solution's value of deriv.  This could be either making deriv much smaller, or incredibly large, or maybe it is fine as is.  I just think we should think a little more about this and make sure it is what we want to do.

talumbau: Yes, this was done in haste to get the small test to pass. I considered a few options here. One thing that seemed reasonable was to have a `permissive` flag (or something similar) that would catch certain exceptions and provide some small (or large) value instead of just raising the exception. In 'real' code, this flag would be False, so that the exception would not be caught. In the "smoke test", we would allow the exception to be caught. Just an idea.

evan-magnusson: Ok, makes sense.  We'll leave this in for now, we can just make an issue and then finalize it later on.

Python/run_simulation.py
```diff
@@ -20,8 +20,6 @@
 import numpy as np
 import cPickle as pickle
 import os
-from glob import glob
```
evan-magnusson: Deleted this two imports, as they are no longer used.

Python/run_simulation.py
```diff
@@ -32,73 +30,15 @@
 
 '''
 ------------------------------------------------------------------------
-Setting up the Model
+Setup
 ------------------------------------------------------------------------
-S            = number of periods an individual lives
```
evan-magnusson: This is now all described in parameters.

evan-magnusson:
In the most recent commit, I moved the folder 'movies' into the python folder.  If we want to get a GIF of the TPI graphs, this folder should be located OUTPUT/TPI folder.  TPI_graphs.py then saves some graphs in the sub-folders of movies, and then we run the py files within them to turn them into GIFs.  We can just leave this for now, it is extremely low priority.

evan-magnusson:
Status: Almost ready.  All the changes have been moved over.  I'm just testing both the run_simulation.py and run_small.py files in different ways to make sure everything works.  Should be done by the end of the day.

evan-magnusson:
I have finished debugging the merged code -- everything runs and mirrors what the code in CURRENT_DEBUGGED_VERSION gave as output.  So, I deleted that folder, and I think it is ready to merge.  I'll comment some of the changes that I put in the code.
@talumbau @rickecon Feel free to go through what I did and let me know if you have any questions/concerns.

Python/dynamic/SS.py
```diff
@@ -103,19 +63,16 @@ def create_steady_state_parameters(a_tax_income, b_tax_income, c_tax_income,
...
-                         d_tax_income]
+    income_tax_params = [a_tax_income, b_tax_income, c_tax_income, d_tax_income]
     wealth_tax_params = [h_wealth, p_wealth, m_wealth]
     ellipse_params = [b_ellipse, upsilon]
-    parameters = ([J, S, T, beta, sigma, alpha, Z, delta, ltilde, nu, g_y,
-                  tau_payroll, retire, mean_income_data] + income_tax_params
-                  + wealth_tax_params + ellipse_params)
-
+    parameters = [J, S, T, beta, sigma, alpha, Z, delta, ltilde, nu, g_y, g_n_ss, tau_payroll, retire, mean_income_data] + income_tax_params + wealth_tax_params + ellipse_params
     iterative_params = [maxiter, mindist_SS]
```
talumbau: typically, I try to keep lines to a max of 80 columns, although I'm often not entirely consistent.

Python/dynamic/SS.py
```diff
@@ -274,78 +280,60 @@ def function_to_minimize(chi_params_scalars, chi_params_init, params, iterative_
...
     ------------------------------------------------------------------------
     '''
 
     if get_baseline:
         # Generate initial guesses for chi^b_j and chi^n_s
         chi_params = np.zeros(S+J)
-        if DATASET == 'REAL':
-            chi_params[0:J] = np.array([2, 10, 90, 350, 1700, 22000, 120000])
-            chi_n_guess = np.array([47.12000874 , 22.22762421 , 14.34842241 , 10.67954008 ,  8.41097278
-                                    ,  7.15059004 ,  6.46771332 ,  5.85495452 ,  5.46242013 ,  5.00364263
```
talumbau: Looks like you were able to get rid of this initial guess for chi_n altogether - is that correct?

Python/dynamic/parameters.py
```diff
@@ -93,40 +189,62 @@ def get_full_parameters():
...
+    PLOT_TPI = True
+    maxiter = 250
+    mindist_SS = 1e-9
+    mindist_TPI = 1e-6
+    nu = .4
     flag_graphs = False
+    #   Calibration parameters
+    # These guesses are close to the calibrated values
+    chi_b_guess = np.array([2, 10, 90, 350, 1700, 22000, 120000])
+    chi_n_guess = np.array([47.12000874 , 22.22762421 , 14.34842241 , 10.67954008 ,  8.41097278
```
talumbau: Oh nevermind my comment above. I see. Yes this is a good place for these parameters.

Python/dynamic/SS.py
evan-magnusson: I didn't even notice it did that.  I changed what was in that list, and so I just copy and pasted over it.  It probably did that in a few places, I can go through and fix that.

PR #102, opened 2015-08-05
Pickle compare utility

talumbau:
- The results of the calculation get stored in pickle files
  This utility function will be used to compare these pickle files
  for different runs (mostly to make sure that changes to the code
  don't produce different results)


talumbau:
committing this additional script as I think it will help with testing of pickle files used throughout.

Issue #107, opened 2015-08-11
Running `run_simulation.py` significantly changes checked in OUTPUT values


talumbau:
I'm interested in making significant structural changes to the `dynamic` package (soon to be renamed `ogusa`?) but I need to know that my changes aren't affecting the answer. So, I checked out a clean version of the repo, and then ran `run_simulation.py`, which was my refactored version of the original script that ran everything. Running this script changes a lot of files currently checked in to the repo. That list of files is as follows:

```
	modified:   Python/OUTPUT/SSinit/ss_init_tpi_vars.pkl
	modified:   Python/OUTPUT/SSinit/ss_init_vars.pkl
	modified:   Python/OUTPUT/Saved_moments/SS_init_solutions.pkl
	modified:   Python/OUTPUT/Saved_moments/labor_data_moments.pkl
	modified:   Python/OUTPUT/Saved_moments/wealth_data_moments.pkl
	modified:   Python/OUTPUT/TPI_K.png
	modified:   Python/OUTPUT/TPIinit/TPIinit_vars.pkl
```
For each of the pickle files, I ran a comparison check using the utility function `pickle_file_compare` in my [recent PR](https://github.com/OpenSourcePolicyCenter/dynamic/pull/102). I checked out another clean version of the repo in a different directory (since the code overwrites what is currently in master). For each pickle file (the clean one in master and the results of running `run_simulation.py`), I captured the output of the `pickle_file_compare`. The output of this utility function is kind of rough, but I cleaned up and condensed some of it. The gist is here:

https://gist.github.com/talumbau/7fef7b31c88a20b9fce1

There are two overall results:

- Many of the values that are different are different by numerical 'noise'. It significantly higher than something like round off error, but I could set either an absolute or relative tolerance for those values and it would be fine. Some examples from the gist:

For the file Python/OUTPUT/SSinit/ss_init_tpi_vars.pkl:
```
'nssmat_init: 
[[ 0.32211472  0.37224071  0.3898083   0.39204884  0.39629503  0.38399747\n   0.28474527]]

[[ 0.32211454  0.37224101  0.38980968  0.39205109  0.39629851  0.38400308\n   0.28475163]]',
```
The first array is one element of `nssmat_init` from running `run_simulation.py`, the second is from master. Not exactly the same, but I could set a tolerance and pass a comparison.

- The values in TPIinit_vars.pkl are wildly different. For example:

```
'rinit: [-0.03865203] [ 0.07479938]
[...]

PR #108, opened 2015-08-19
changed w and r to wb and rb in pickle code

evan-magnusson:
See issue #107 .  The lines involving pickling have been modified.

@talumbau


talumbau:
what's the deal with `TPI_K.png`? Very different, right? Will check pickle files shortly.

talumbau:
Worked! The TPIinit_vars.pkl file reads in with the expected values now. I don't know how to explain the change in the `TPI_K.png` file, but it's nice that github has added the feature so you can look at them side by side. Once that is resolved to @evan-magnusson's, @jdebacker's, @rickecon's, and/or @isaacswift's satisfaction, I recommend merging this PR.

evan-magnusson:
This is the  correct TPI_K.png.  The other one must have been from a simulation where I didn't let it finish, and then I think I let it finish but didn't create the new graphs.

PR #113, opened 2015-08-28
Refactor IO code to abstract out "OUTPUT"

talumbau:
- allow one to specify a different output directory


talumbau:
The purpose of the PR is to begin abstracting out the hard-coded paths for reading/writing pickle files. Now that we can get the same answers on all platforms, we need to move to a style of computation like this:
```
reference inputs + code -> outputs
compare outputs to reference outputs -> if difference is less than TOL, tests PASS
```
Right now, the reference inputs, reference outputs, and code outputs are all mixed together. By separating these inputs and outputs, we can then do more invasive changes to the code and ensure that the answer remains the same.

talumbau:
I think this is in reasonable shape. I would prefer a resolution on https://github.com/OpenSourcePolicyCenter/dynamic/issues/115 before merging.

talumbau:
This now makes use of the file comparison features and will run the full SS+TPI by running `py.test`. (this takes about 40 minutes, so for Travis CI, we skip that test).

talumbau:
Merging this doesn't impact running the `run_simulation.py` script, but it does provide the option to put all output in a different directory. Just FYI

cc @rickecon @jdebacker @evan-magnusson @isaacswift @MattHJensen

Issue #115, opened 2015-08-31
What is the proper error metric for the results of running the model?


talumbau:
Now that we have resolved the issue with the pickle files, I updated the comparison functions with this PR:

https://github.com/OpenSourcePolicyCenter/dynamic/pull/114

and I used that function in order to see the level of error when just running the basic simulation. The test file looked like this:

https://gist.github.com/talumbau/47e9bd02ad9698bcfe04

I output the results to a file, which you can see here:

https://gist.github.com/talumbau/657f208367edd024848e

Most of the values have an error within an acceptable range. There are some concerning error levels though. In particular, these error rates are troubling:

in ss_init_vars.pkl:

```
diff for bssmat_s is 0.00147395404773 which is NOT OK
diff for factor_ss is 0.754183860641 which is NOT OK
diff for bssmat_splus1 is 0.00162147841888 which is NOT OK
```

in SS_init_solutions.pkl:

```
diff for solutions is 0.754183860641 which is NOT OK
```

in TPI_init_vars.pkle:

```
diff for b_mat is 0.00362436350586 which is NOT OK
diff for c_path is 0.0047934971713 which is NOT OK
```

I will run these on a Windows VM just to make sure there are no platform issues, but I don't have a good sense of if these error levels are OK. To be clear, this is a comparison of all of the pickle files in master versus the state of the pickle files after running run_simulation.py. This is an absolute measure of the difference, so maybe in some cases a relative error metric would be better. Any thoughts on this would be appreciated. cc @evan-magnusson @rickecon @jdebacker @MattHJensen

talumbau:
for some reference:

`bbsmat_s` is a numpy array of shape (80, 7) with a max of 87.520727775315791, mean of 8.0493978367920835 and min of 0.0

`bbsmat_s` is a numpy array of shape (80, 7) with a max of 87.520727775315791, mean of 8.0493978367920835 and min of 0.0

`factor_ss` is a scalar equal to 179436.48618511992

`bssmat_splus1` is an array of shape (80,7), max of 89.452041776928212, min of 0.022723523254224953 and mean 8.2734565816308017

`solutions` is an array of shape (1124,), max of 179436.48618511992, min of 0.011653966176984784 and mean of 163.89156041412156

`b_mat` is an array of shape (240, 80. 7), max of 95.246160048505004, min of 0.0, mean of 7.398549236956633

`c_path` is as follows:

```
>>> np.max(c_path)
2.898146087925614
[...]

PR #116, opened 2015-09-09
Update comparison functions

talumbau:
- add 'relative' flag
- if all a and b values are smaller than EPSILON, return True


talumbau:
Merging this for continued investigation of issue #115

PR #123, opened 2015-09-24
Pep it

talumbau:
Ran all of the Python through the autopep8 tool. I am not necessarily advocating that we enforce PEP8 in this code base right now, but using the tool has allowed me to easily look at warnings that show global variable usage, or even unintentional use of variables grabbed in from the global namespace. This will aid in tracking down issues, since those kind of warnings are pretty much the only ones left in the code.


talumbau:
Note that this is a very "light" touch with the autopep8 tool. The tool was invoked with zero `--aggressive` flags, so PEP8 violations like line length are still allowed. But it does increase readability significantly, in my view.

talumbau:
this PR passes the 'big' test, which is run like this:

```
py.test -s dynamic/tests/test_sstpi.py
```

PR #130, opened 2015-10-02
Added initial Python scripts in folder dynamic/Python/firms that solve the steady-state and transition path equilibria.

rickecon:
The files main.py, ssfuncs.py, and tpfuncs.py solve for the steady-state and transition path equilibria for a simple overlapping generations model with two industries, two consumption goods, and S-period lived agents. We still need to clean up the code and improve the commenting. But this is the foundation we will build from.


jdebacker:
Great.  Will start contributing to this now.

PR #133, opened 2015-10-02
clean out unused tp functions

jdebacker:
This pull request:
- Deletes alternative, unused versions of get_Ympath, get_Kmpath, get_Lmpath
- Deletes convergence method
- Renames TPI function as TP- it is called after fsolve to get all the other endogenous variables along the time path (TPI_fsolve just returns rpath, wpath).  We might want to consolidate this later - some redundancy between TP and TPI_fsolve.

### DIFF
--- a/Python/firms/main.py
+++ b/Python/firms/main.py
@@ -35,23 +35,23 @@ T           = integer > S, number of time periods until steady state
 alpha       = scalar in (0,1), expenditure share on good 1
 c1til       = scalar > 0, minimum consumption of good 1
 c2til       = scalar > 0, minimum consumption of good 2
-cmtilvec    = [2,] vector, minimum consumption values for all goods
+cm_tilde    = [2,] vector, minimum consumption values for all goods
 beta_ann    = scalar in [0,1), discount factor for one year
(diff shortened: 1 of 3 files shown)


jdebacker:
Second commit to the pull request updates names of numpy arrays of variables from things like "epsvec" to things like "epsilon".

rickecon:
The nice thing about epsvec instead of epsilon is that the name communicates the idea that the object is a vector of epsilons. This is more intuitive as we add more industries and those vectors become longer.

rickecon:
And we are correctly removing the bisection method, referred to above, from this code. The bisection method does not reliably work with more than one variable (see Kopecki, http://www.karenkopecky.net/Teaching/eco613614/Notes_RootFindingMethods.pdf).

jdebacker:
Re names of arrays of variables.  My thought is that the "vec" suffix is redundant.  Because we are going to write the code such that it is generalizable to SxJxMxIxT, everything will be done with array operations and so it is implied that the length along any dimension may be greater than one.

rickecon:
I agree @jdebacker.

PR #135, opened 2015-10-03
[WIP] SS functions vectorized

jdebacker:
SS functions have been vectorized and run, but SS sol'n differs from non-vectorized code.  This suggests an array broadcasting error that needs to be traced out.


jdebacker:
SS sol'n now same in vectorized and non-vectorized code (error was in the alpha parameter vector - simple typo).  Started to work on the time path code in the main file.

rickecon:
@jdebacker Do you want to do the PR now or should we wait until TPI is working as well with the newly vectorized functions?

PR #141, opened 2015-10-08
consolidated SS and TP functions

jdebacker:
This PR consolidates functions that were common to the SS and TP solutions.  These are now in a separate script - firm_funcs.py.  

This will help avoid errors as we update to the firm problem since it minimized the number of times equations need to be changed.


jdebacker:
Additional commits to this PR:
- Replace the nonlinear solver used to solve for the time path of output with a linear solver.  This increase speed of the time path solution by about 50%
- Allows the user to choose to calibrate the consumption subutility function and the firms to the parameter values of Fullerton and Rogers (Brookings, 1993).  This model has 17 consumption good and 19 production industries.

PR #154, opened 2015-10-28
Added txfunc.py, which estimates effective tax functions

rickecon:
Added txfunc.py, which estimates effective tax functions as a function of labor income and capital income. This file has an efficient structure of looping through 10 years of cps-puf csv files, and within each file estimating functions for each age. This file also has some nice visualization tools to make sure the fit of the function looks right with the data.

### DIFF
--- a/Python/microtaxest/txfunc.py
+++ b/Python/microtaxest/txfunc.py
@@ -1,0 +1,267 @@
+'''
+------------------------------------------------------------------------
+This script reads in data generated from the OSPC Tax Calcultor and
+the 2009 IRS PUF.  It then estimates tax functions tau_{s,t}(x,y), where
+tau_{s,t} is the effective tax rate for a given age (s) in a particular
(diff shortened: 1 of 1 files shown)


jdebacker:
Looks good.  A couple small comments:
1) Loops over the Tax Calculator output, not the CPS-PUF (which is just one year of data)
2) Should set the limits for the AETR (used to determine outliers) from the Tax Calculator data to be functions of the user entered tax parameters.

rickecon:
@jdebacker. I edited my comment to have the correct data reference. Also, note that the functions do not yet fit the data well. We still have some work to do on functional forms. But I wanted to get the visualization tools and the script structure to you and @chrisrytting so you could play with the estimation.

rickecon:
Note that I need to change the functional form of the ratio of polynomials to be the following:

tau(x,y) = (maxt - mint) * (P(x,y) / P(x,y) + k) + mint

PR #157, opened 2015-11-04
Make dynamic-->ogusa

talumbau:
This PR changes the name of the Python package from `dynamic` to `ogusa`, the agreed upon name from the discussion some time ago. Follow-on PRs to this will create the conda recipe, which is the first step to having a way to deploy this package in our worker node system. @rickecon @jdebacker the script `run_simulation.py` has been minimally revised to handle this change. cc @MattHJensen


rickecon:
@talumbau . All Travis CI checks have failed.

talumbau:
yep, just pushed an additional commit to fix. We'll see if that fixes everything.

PR #158, opened 2015-11-05
Updated tax function estimation script with timing and pickle

rickecon:
I updated the Python/microtaxest/txfunc.py function with timing capability as well as a pickle command to save the parameter array. This pickle of parameters will be passed in to the OG-USA model.


rickecon:
For the txfunc.py script, I still need to write some functionality that intelligently fills in missing portions of the parameter array.

PR #159, opened 2015-11-06
Conda recipe

talumbau:
This PR adds a few files to the top level directory:

MANIFEST.in
versioneer.py
setup.py
setup.cfg

These are at the top level but only pertain to the Python/ogusa package. This seemed to be the only way to get versioneer working properly, which is a nice way to tie versions to git tags. This is what we do with `taxcalc` so that one can easily bump versions as new features/bug fixes are added.


rickecon:
@talumbau . Travis CI checks have failed.

talumbau:
thanks @rickecon It will likely fail several more times this morning as I work through the CI issues. I'll be changing test and install scripts, so it might take me a few tries to get everything right.

talumbau:
verified that this passes the more stringent `Python/tests/test_sstpi.py` so I see no reason not to merge.

PR #163, opened 2015-11-11
Updated documentation, average income, and blank age data in txfunc.py

rickecon:
This PR has 3 commits for the function Python/microtaxest/txfunc.py that estimates continuous tax liability functions given data from Tax-Calculator. The first commit updated the documentation of the script. The second commit calculated average income for each year from the Tax-Calculator data. And the third commit introduced code that linearly interpolated estimated parameter values for ages without sufficient data.


rickecon:
The output of this function is TxFuncEst.pkl, which is a dictionary containing 3 objects. The first object is a numpy array of parameter estimates for the tax functions by age (s) and year (t). The second object is a vector of average total income values for each year. The third object is the elapsed time for the computation of the tax functions.

PR #165, opened 2015-11-15
New tax funcs

jdebacker:
This PR:
- adds more robust solution to household problem (using sol'ns from other types to update initial guesses)
- able to solve SS with taxes = 0
- HH problem solve with calibrated tax functions, but GE not converging in this case


rickecon:
@jdebacker Is this ready to merge? I would suggest that we merge it, and then keep working on the non-steady-state solution. It is nice to have this more robust steady-state solution.

jdebacker:
@talumbau I want to merge this, but I Travis CI build fails.  I don't want to affect files you are working on, but these scripts I want to merge run fine on my machine.  Can you tell me if this is ok to merge?

talumbau:
@jdebacker looks like some good updates to the code, but the test failure is genuine. It might just be that we need to update the test harness. For Travis CI, we use run_small.py in the Python/ directory. It might just need to be updated. The idea is that we just want to make sure that one could do `python run_small.py` on master and have it work. If you click the bottom red 'X' next to your latest commit:

screen shot 2015-11-20 at 12 51 37 pm

it will take you to the Travis CI page where you can see the actual error message.

rickecon:
@jdebacker @talumbau . It might be that we didn't update the get_reduced_parameters() function in parameters.py. If my understanding is correct, Travis CI depends on that small version of the model. it reduces J=2 but keeps S=80. That might be causing an issue.

jdebacker:
Travis CI checks pass, so I'll merge.  But model note that all model constraints are not met in the small parameter case (S=10, J =2).  This shouldn't matter - the check it just that the code runs without error.  parameters are not calibrated to this limited model so convergence to a tight tolerance isn't really relevant.

PR #172, opened 2015-12-08
Update dynamic code to use MTR functions

jdebacker:
This PR updates the OG-USA code to allow for separately estimated MTR functions.  

It also stream lines the code a bit so that running run_simulation_from_json.py will take create the microdata (with MTRs) , estimate the tax functions, and solve the model in the SS and along the time path.


talumbau:
OK @jdebacker I will continue to put updates/refactors in this PR. On my local machine, I have Amy's PR with the `capital_mtr` function for `taxcalc` and I have the PUF, so I can import `ogusa`. After running for a while, the output I get is:

```
Tax function estimation time:  36  min,  36.5  sec
```

So I think I'm on the right track in running everything.

jdebacker:
@talumbau The commit above related to txfunc.py should help integrate it into the parameters.py file by creating a tax_func_estimate function.  

These commits also fix:
1) get_micro_data.py - updated to grab combined rates for all years -  prior version just did this for 2015.
2) the calculation of the labor MTRs in txfunc.py (type created wrong ratios when doing weighted avg of mtr on wages and mtr on self-emp income.

PR #176, opened 2015-12-17
add script to postprocess macro_vars.pkl files

talumbau:


talumbau:
these are a couple of helper scripts to help TaxBrain run the OG-USA package.

PR #178, opened 2015-12-22
[WIP] Updates passing of tax parameters, labor supply elasticity

jdebacker:
This PR does the following:

1) Updates all ogusa files that handle tax function parameters to pass them in 3 large arrays (for the ETR, MTR Labor, and MTR Capital functions) rather than 30 smaller arrays.  
2) Adds a script, elliptical_u_est.py, that fits an elliptical utility function to a constant Frisch elasticity function.
3) Updates parameters.py take as a parameter the Frisch elasticity of labor supply and then fits the elliptical utility parameters to this by calling elliptical_u_est.py.
4) Updates the model parameter sigma to match the literature and provides citations (in comments) of the sources of delta, alpha, beta, and the Frisch elasticity.


jdebacker:
Note that this is not ready to merge.  The model does not converge with the new labor supply elasticity parameters. Our calibration of the chi_n and chi_b parameters needs to be updated given the new parameter values and tax functions.  It is expected that the model will converge once these parameters are updated.

jdebacker:
I think with this last commit we are good to merge this PR.  The model should run through and produce sensible output.

talumbau:
OK, I am re-running based on your changes. Let me look at any issues related to Travis CI as well before we merge. I don't think it will be difficult to get Travis CI passing again.

PR #186, opened 2016-01-08
frisch constrainted with validation in metadata

talumbau:
- 0.1 <= upsilon <= 0.8 for now

### DIFF
--- a/Python/execute.py
+++ b/Python/execute.py
@@ -12,7 +12,8 @@ import ogusa
 ogusa.parameters.DATASET = 'SMALL'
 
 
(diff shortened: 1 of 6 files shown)


jdebacker:
@talumbau We need to pass (and constrain to the range noted here) the parameter "frisch", which is the Frisch elasticity of labor supply and right now hard coded in parameters.py.  Given this parameter, we find upsilon by fitting the elliptical utility function to a constant Frisch elasticity function, but the user will enter the Frisch elasticity, not upsilon.

talumbau:
@jdebacker thanks, fixed in this PR. please have a look and let me know.

jdebacker:
Looks good to me.

PR #193, opened 2016-01-28
Allow for analytical mtr functions

jdebacker:
This pull request creates the option to use either the "analytical" marginal tax rate functions (i.e., those derived from the total tax function implied by the ETR function) or the marginal tax rate functions that are unrelated to the ETR functions.

The option is controlled by a boolean, "analytical_mtrs" that is entered in the key word arguments for the model run in run_ogusa.py.  When set to True, the analytical MTRs are used. Else the MTRs unrelated to the ETR functions are used.


Python/ogusa/tax.py
```diff
@@ -197,44 +197,83 @@ def MTR_capital(r, b, w, e, n, factor, mtry_params):
...
+        denom = (A*(x**2)) + (B*(y**2)) + (C*x*y) + (D*x) + (E*y) + F
+        Lambda = num/denom
+
+        d_num = (2*B*y + C*x + E)*F
+        d_denom = ((A*(x**2)) + (B*(y**2)) + (C*x*y) + (D*x) + (E*y) + F)**2
+        d_Lambda = d_num/d_denom
+
+        tau =  (max_y-min_y)*Lambda + (x*(max_x-min_x) + y*(max_y-min_y))*d_Lambda + min_y
+
+    if not analytical_mtrs:
```
talumbau: could probably just do `else:` here

Python/ogusa/tax.py
```diff
@@ -250,40 +289,80 @@ def MTR_labor(r, b, w, e, n, factor, mtrx_params):
...
+        denom = (A*(x**2)) + (B*(y**2)) + (C*x*y) + (D*x) + (E*y) + F
+        Lambda = num/denom
+
+        d_num = (2*A*x + C*y + D)*F
+        d_denom = ((A*(x**2)) + (B*(y**2)) + (C*x*y) + (D*x) + (E*y) + F)**2
+        d_Lambda = d_num/d_denom
+
+        tau =  (max_x-min_x)*Lambda + (x*(max_x-min_x) + y*(max_y-min_y))*d_Lambda + min_x
+
+    if analytical_mtrs:
```
talumbau: is this right? At line 292 you have `if analytical_mtrs`, but now here again you have `if analytical_mtrs`?

Python/ogusa/tax.py
jdebacker: Yes - will update PR.

Python/ogusa/tax.py
jdebacker: This is not right.  Fixing now and will update PR.

talumbau:
@jdebacker I had a comment on one line and a question about another one. I will submit a follow-up PR to prevent duplication in `execute.py` and `execute_large.py`. I should have fixed that some time ago.

PR #195, opened 2016-02-01
Update SS.py to include MTRs in output

jdebacker:
This PR does 2 things:
1) It allows the user the option to see MTRs by age and lifetime income group in the the SS distribution.
2) It updates macro_output.py to output the pct difference from the SS (prior to this PR, the "SS" results were just from the end of the transition path).  This would have no effect if the TP converges properly, but since there have been cases where the TP did not converge, it will be helpful to output the true SS values here.


rickecon:
This is the base that @chrisrytting and I will use to debug the ETR and MTR tax function estimation routines.

PR #196, opened 2016-02-09
Fixes for tax function outliers

jdebacker:
It was observed that that tax functions for specific age-years could not be fit well, despite a large number of observations.  This PR replaces poor fitting tax functions via a linear interpolation.

A few minor changes to other scripts are included in this PR.


Python/ogusa/macro_output.py
```diff
@@ -55,9 +55,9 @@ def dump_diff_output(baseline_dir, policy_dir):
     
     ## Load SS results
     ss_policy_path = os.path.join(policy_dir, "SSinit", "ss_init_vars.pkl")
-    ss_policy = pickle.load(open( tpi_macro_vars_policy_path, "rb" ))
+    ss_policy = pickle.load(open( ss_policy_path, "rb" ))
```
talumbau: this change looks correct, because the comment says "Load SS results". Does that mean it was loading the incorrect file before?

Python/ogusa/macro_output.py
```diff
@@ -55,9 +55,9 @@ def dump_diff_output(baseline_dir, policy_dir):
...
     ss_baseline_path = os.path.join(baseline_dir, "SSinit", "ss_init_vars.pkl")
-    ss_baseline = pickle.load(open( tpi_macro_vars_baseline_path, "rb" ) )
+    ss_baseline = pickle.load(open( ss_baseline_path, "rb" ) )
```
talumbau: same comment as above for this change.

Python/ogusa/macro_output.py
jdebacker: That's right @talumbau, before the change that wrong file was loaded, which would cause an error in calculating the SS values.

talumbau:
ok great, +1 from me.

PR #202, opened 2016-02-25
Baseline SS results used for reform

jdebacker:
This PR makes two fixes to the model solution of the reform policies:
1) It uses the model adjustment factor (which scales model income so that it matches income from the PUF) found in the baseline SS solution to adjust model income in the reform as well.  
2) It ensures that the initial period capital stock for the reform is the same as that for the baseline.

### DIFF
--- a/Python/execute.py
+++ b/Python/execute.py
@@ -12,7 +12,7 @@ import ogusa
 ogusa.parameters.DATASET = 'SMALL'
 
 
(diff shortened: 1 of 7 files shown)


rickecon:
@talumbau . I think this pull request is ready to merge, but Travis CI checks are failing. I know @jdebacker updated the kwargs that must be passed in both the reform and the baseline. Can you help us fix the "small" run in Travis CI so that these changes are incorporated there as well?

PR #206, opened 2016-03-11
Representative agent model

jdebacker:
This PR reduces the complexity of OG-USA.  It replaces heterogenous lifetime income groups, with a single, representative agent for each cohort.  It also replaces tax functions where marginal and average tax rates vary over current income with ETRs and MTRs that are constant across age and income.


jdebacker:
FYI, the script in this PR, get_micro_data.py is not compatible with the current Tax Calculator. Once tax calculator PR #638 is merged in, it will work.

CC: @talumbau, @rickecon

MattHJensen:
@jdebacker, I do not expect https://github.com/open-source-economics/Tax-Calculator/pull/638 to be merged soon because it would require deploying a new datafile to taxbrain right now, which seems risky. Do you depend on the entire PR? What you really need is just @martinholmer's latest, commit, https://github.com/martinholmer/Tax-Calculator/commit/d9ce73ae8d7bf3b2b046b5b53e8f7a222497787f, right? In which case we should have that be a separate PR to tax-calculator.

MattHJensen:
> What you really need is just @martinholmer's latest, commit, martinholmer/Tax-Calculator@d9ce73a, right? In which case we should have that be a separate PR to tax-calculator.

@jdebacker, even that is slightly risky because it requires a new version of tax-calculator on the webapp, is there any easy way to make it so that this PR works w/o any changes to tax-calculator? 

cc @talumbau @martinholmer

martinholmer:
Matt said:
> I do not expect open-source-economics/Tax-Calculator#638 to be merged soon because it would require deploying a new datafile to taxbrain right now, which seems risky. 

What exactly about pull request open-source-economics/Tax-Calculator#638 is risky?

@MattHJensen @jdebacker @talumbau

MattHJensen:
@martinholmer, as @talumbau requested:

> because this PR would make Tax-Calculator master non-backwards compatible with the existing PUF, can we put a one week hold on this? There are many balls in motion in regards to OSPC work right now and this would be a bad time to introduce a change into master that would break existing users of the current PUF.

We are still inside that one week request from @talumbau, and in fact I think that was probably not asking for quite enough time.

jdebacker:
@talumbau @martinholmer @MattHJensen  Yes, let me replace one variable with another and it'll be good to go.  We can update later.

MattHJensen:
Thanks @jdebacker

jdebacker:
@talumbau, I'm getting a Travis CI failure.  It appears to be because the test doesn't have an up-to-date pickle of the tax functions.  I just pushed a new one to the repo, but maybe it's not what the test is looking for. Can you let me know what file name and directory the pickle should be in?  As best I can tell it's looking for /OG-USA/Python/TxFuncEst_baseline.pkl - but that's the one I updated and I still get the build failure.

Thanks!

talumbau:
@jdebacker the test failure is the result of calling `get_full_parameters` with `baseline=False`. This causes the function to read, this case, the *policy* pickle file, instead of the baseline. Can you `Python/TxFuncEst_policy.pkl` as well? I think this will solve the problem.

PR #212, opened 2016-03-30
add exceptions to SS and TPI solutions

jdebacker:
This PR adds exceptions, noting when an equilibrium has not been found.

Messages generated to from these exceptions need to be added to:

SS.py on lines: 689, 707, and 834

TPI.py on line 674

This PR should not be merged until these are updated.

### DIFF
--- a/Python/ogusa/TPI.py
+++ b/Python/ogusa/TPI.py
@@ -670,6 +670,8 @@ def run_time_path_iteration(Kss, Lss, Yss, BQss, theta, income_tax_params, wealt
         print '\tIteration:', TPIiter
         print '\t\tDistance:', TPIdist
 
+    if TPIiter >= maxiter:
+        RAISE EXCEPTION - Transition path equlibrium not found
 
     return winit[:T], rinit[:T], T_H_init[:T], BQinit[:T], Yinit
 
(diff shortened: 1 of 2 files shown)


jdebacker:
Updated TPI.py to include a couple of exceptions raised by @rickecon

talumbau:
@jdebacker @rickecon the ball is in my court on this, right? If we get these exceptions in, then we should be good to release. Do you agree? Or do you have changes coming soon so that we should hold off on a release?

jdebacker:
@talumbau yes, we can use your help on this.  Let me close this PR since the code has changed so much.  ShallI open a PR like this that you add to or can you take what is here (which isn't much of a change) and go with that?  Let me know.  I can get to this in a couple hours.

Also, @rickecon may have a quick fix to some of the constraint punishers. @rickecon can those be included quickly for the new release?  Or rather take a bit more time to do some more testing?

Python/ogusa/TPI.py
```diff
@@ -832,7 +834,10 @@ def TP_solutions(winit, rinit, T_H_init, BQinit2, Kss, Lss, Yss, BQss, theta, in
     eul_laborleisure = euler_errors[:, S:, :].max(1).max(1)
 
     print 'Max Euler error, savings: ', eul_savings
-    print 'Max Euler error labor supply: ', eul_laborleisure
+    print 'Max Euler error labor supply: ', eul_savings
```
talumbau: @jdebacker this is a typo, correct? I think the original line:

```
print 'Max Euler error labor supply: ', eul_laborleisure
```

should stay. Do you agree?

jdebacker: @talumbau I agree.  But where is this?  If I look at line 643 of TPI.py in open-source-economics/OG-USA:master, it looks correct.

jdebacker:
Replaced by PR #216

PR #213, opened 2016-04-01
Update outlier identification in txfunc.py

rickecon:
1) This PR updates the outlier identification in the find_outliers() function in txfunc.py. In particular, I look at SSE/N, where SSE is the sum of squared errors and N is the number of observations in a particular age and year. This provides better outlier identification than the previous method that only tracked SSE.

2) I also did extensive updating of the commenting docstrings in the txfunc.py module. It still needs some more work, but the pattern is there.


rickecon:
Here is a plot of the adjusted sum of squared errors (SSE/N) for estimating the marginal tax rate of labor income (mtrx) functions for each age (s) in each year (t).
sse_mtrx
After sweeping out that one spike in the adjusted SSE of the mtrx tax function for 48-year-olds in 2016, the following three outliers are identified, which I think are two in 2020 and one in 2025 (or 2018).
sse_mtrx_noout1
Once we interpolate those outliers with replace_outliers(), we have the following result of adjusted SSEs with no apparent outliers.
sse_mtrx_noout2

jdebacker:
Great - scaling by the number of observations is the right way to measure goodness of fit.  Merging.

PR #217, opened 2016-05-04
Updated demog.py

rickecon:
This PR updates demog.py. There were some errors in the previous version of this module. 
- This fixes the calculation of the adjusted immigration rates.
- It outputs an S x T+S matrix of original immigration rate vectors and adjusted immigration rate vectors. 
- The law of motion for the time path of the population distribution is fixed.
- It adds "*/OUTPUT/*" and "puf.csv" to the .gitignore file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,5 @@ dist/*
 *.m~
 *.sublime*
 .DS_Store
+*/OUTPUT/*
+puf.csv
--- a/Python/ogusa/demog.py
+++ b/Python/ogusa/demog.py
@@ -6,6 +6,7 @@ model
 This module defines the following function(s):
     get_fert()
     get_mort()
+    pop_rebin()
     get_imm_resid()
     immsolve()
     get_pop_objs()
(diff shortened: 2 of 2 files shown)


rickecon:
@talumbau @jdebacker Could this PR be failing TravisCI because of the changes I made to .gitignore (i.e., "*/OUTPUT/*" and "puf.csv"? Because otherwise, the only changes were to demog.py which is not currently integrated with OG-USA.

rickecon:
@talumbau The TravisCI failure seems to be with something I did wrong in merging the changes in my demog branch of my OG-USA fork into the master branch of my OG-USA fork. How do I fix this?

talumbau:
In this PR, the function `get_pop_objs` returns 7 values now, but other uses of the function throughout the code base depend on the function returning 6 values. I would search the code for all instances of `get_pop_objs` and see how you would prefer that these calls are modified now that the function returns 7 values. When I click on `Details` for the failed Travis CI build, it looks like at least one instance is on line 356 of parameters.py:

screen shot 2016-05-04 at 1 55 30 pm

talumbau:
oh some of the changes are the result of a bad merge? Hmm.. that one is difficult. I would have to understand what changes are intentional and what are not. we might need to set up a quick hangout to fix.

rickecon:
I am closing this PR and submitting a new one.

PR #221, opened 2016-05-19
New demographics.py

chrisrytting:
Here we've replace the old demographics.py with a new demographics.py, which is more thoroughly commented, more effectively modularized, and intended to more accurately make calculations necessary to OG-USA (e.g. immigration rates).


jdebacker:
@chrisrytting, Changes look good.  Before merging, let's workout issues with Travis CI built fail.  Current issue is an import of demog in execute_small.runner().  

FYI - if you click on "Details" next to the Travis CI build failed message, you will probably see enough detail to work out any other issues.

talumbau:
I can't judge the merit of code itself, but it does seem good to only have one demographics module, instead of `demog.py` and `demographics.py`. The only thing is that we still have code around that imports both. That is why the one test is failing. If you could remove the imports of `demog` from these two files, I think that will fix the failing test.

https://github.com/open-source-economics/OG-USA/blob/master/Python/execute.py

https://github.com/open-source-economics/OG-USA/blob/master/Python/execute_small.py

chrisrytting:
@jdebacker @talumbau Alright I've fixed the issue in execute_small.py by changing every instance of demog to demographics and it seems as if the checks have passed, so I would imagine we're good to go now.

talumbau:
we don't have a great mechanism for this yet, but `execute_small.py` is meant to "mirror" the behavior of `execute.py`, so changes in one should happen in the other. The point of `execute_small.py` is to make it so that we can quickly test a code path that is in use for the web application. So, if you change one, you should really change the other to make sure they stay in sync. Until we come up with a better solution, the only alternative to this arrangement was to have the test take 2.5 hours to run, which seemed bad. So would you mind making a similar change in `execute.py` so they stay in sync. Apologies for the confusion.

PR #222, opened 2016-05-19
Changed instance of demog to demographics in execute.py.

rickecon:
I fixed my "oops" of merging @chrisrytting 's PR too quickly. There was only one instance of demog remaining in execute.py. That is now fixed to demographics.py.


talumbau:
ok great thanks @rickecon!

PR #223, opened 2016-06-06
Population dynamics

jdebacker:
This PR is a WIP, the work in progress label will be taken down when all checks are complete.  

The PR begins to bring back rich population dynamics, including immigration rates.  PR #221 had a few issues this addresses. In particular, this PR updates demographics.py, the capital market clearing equation (household.get_K), and the aggregate investment equation (firm.get_I) to account correctly for the inflow and outflow of capital due to immigration.  

The code has passed checks for the SS (resource constraint error of -8.20454815198e-14). However, the transition path still gives unacceptably large resource constraint issues.  Adjustments to the timing of the transition equations should fix this.  At that point the WIP label will be removed.


jdebacker:
Still working on this, but with latest commits (6/24), seems that error is with immigration/pop somewhere, unlikely with general timing in TPI.  Errors with non-constant taxes and constant pop gives max error of -6.33573194e-11.  This suggests something with pop dynamic.  

Also, with nonconstant pop dynamic, errors much smaller after "fixper" (e-12 except for period T) and larger before (e-5 before fixper).

Next will look back at demographics.py and all references to imm_rates in TPI code.

jdebacker:
Max RC error now 1.06316067e-11.  Satisfied with this.  Will do full run of baseline and reform and confirm results before taking off WIP label.

jdebacker:
RC errors now confirmed to be 1e-10 or smaller for all periods.  Latest commits update SS that reform TP moves towards- it had mistakenly taken the baseline SS.  

Due to the updated population dynamics we need a longer timepath.  T should be set to 3*S or greater to ensure SS obtained along time path.

jdebacker:
@talumbau , can you give me a hand with this Travis CI fail?  It has to do with a call to the Tax Calculator specifying a variable that no longer exists.  The call to that variable is in the taxcalc code that is called by the test.  It seems that the code is outdated, but that it's trying to call a recent PUF.  But I'm not sure where the taxcalc code that is called resides or how to change it.

Thanks!

rickecon:
I think the message in commit d8dd577 is a little misleading. @jdebacker @talumbau

jdebacker:
Thanks @talumbau !  I think this is PR is ready to go.

rickecon:
I don't think @jdebacker got my joke.

talumbau:
Correct.

jdebacker:
Ah - yes, pretty epic typo there!

PR #225, opened 2016-06-28
Update test PUF data

talumbau:
update test PUF data with puf91taxdata.csv.pz for version 0.6.4 of `taxcalc`


talumbau:
@jdebacker if we merge this, we should be able to fix issues in #223

PR #228, opened 2016-07-18
fixes to SS.py for reform runs

jdebacker:
This PR fixes a bug with the SS solution for reform policies.  The resource constraint was not satisfied because of an error with how the model income factor was being applied.

### DIFF
--- a/Python/ogusa/SS.py
+++ b/Python/ogusa/SS.py
@@ -239,7 +239,7 @@ def euler_equation_solver(guesses, params):
     return list(error1.flatten()) + list(error2.flatten())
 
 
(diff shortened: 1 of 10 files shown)


jdebacker:
This PR adjusts tax functions to limit potential for multiple equilibria.  A better solution should be added (by constraining the parameters for the tax functions in estimation), but this PR will improve functionality and be an intermediate step.

jdebacker:
As a note, these commits leave a lot of extraneous code in parameters.py and a few lines in SS.py. These are left in for help debugging future issues.

Also, note that the values in parameters.py are set to help the model run quickly, but are not what we would want the parameters to be set at for policy analysis.

PR #229, opened 2016-07-27
Calibration

jdebacker:
This PR updates the calibration routine.  The central script is calibrate.py, which calls labor.py and wealth.py to get the data moments for labor and wealth, respectively.  The wealth moments chosen have been updated to reflect cross-sectional moments that are more widely known.  In addition, the scripts have been restructured to be more easily read and allow for easier modification.

An optimal weight matrix and the calculation of standard errors will be added through a future PR.

### DIFF
--- a/Python/execute.py
+++ b/Python/execute.py
@@ -229,7 +229,7 @@ def runner_SS(output_base, baseline_dir, baseline=False, analytical_mtrs=True, a
     CALL CALIBRATION here if boolean flagged
     ****
     '''
-    calibrate_model = True
+    calibrate_model = False
     if calibrate_model:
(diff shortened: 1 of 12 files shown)


jdebacker:
Note: Travis CI will fail if boolean for calibration = True.  Will turn back off when all work on this PR finished.

rickecon:
@jdebacker is this ready to be merged? Your last comment seemed to suggest that this PR is a WIP. Let me know.

jdebacker:
I think we are ok to merge.  One just has to be sure (as was always the case, but more now) that all the keywords arguments and parameters are set to what they want them to be when running.

Issue #230, opened 2016-07-28
Create income.py interpolant for general S and J


rickecon:
I am currently working on making the income.py module robust to general ages and ability types S and J. The module currently only works with S=80 and J=7. Because the resulting 80 x 7 ability matrix is smooth, we can simply use scipy.interpolate.griddata() to interpolate this.

I currently have a function written that takes general S and J midpoints of income distribution bins to interpolate ability matrix values. The following 3D plot is the original 80 x 7 ability matrix.
![ability_orig](https://cloud.githubusercontent.com/assets/4317690/17202652/26371f0a-5460-11e6-903d-1c8d61d07f57.png)
A version that I interpolated using the above described griddata() function with S=20 and ability distribution midpoints of np.linspace(0.15, 0.95, 10) gives the following interpolated plot.
![ability_interp](https://cloud.githubusercontent.com/assets/4317690/17202700/67c93f16-5460-11e6-802f-6490c47647df.png)
This looks right. I now need to incorporate this function into income.py and into parameters.py, which calls the get_e function from income.py. @jdebacker

rickecon:
I was originally trying to use some bivariate kernel density estimation code to do this. But that approach is better when the data need to be smoothed. This original data in the first plot above is already smooth. So a simple interpolator works best.

rickecon:
By the way, the second picture above uses the "...method='linear'" option. The "cubic" option, for cubic spline, gives too many curves. See the following graph. The linear interpolation method above is clearly better.
![ability_interp_spl](https://cloud.githubusercontent.com/assets/4317690/17202868/8d458c1c-5461-11e6-808b-46f9f4e76ba4.png)

rickecon:
@jdebacker , @kerkphil : I now have this working for calculating the steady-state with S=20 and J=1. It does not converge in the steady-state for anything lower than S=20. Now I just need to make sure that TPI works.

PR #231, opened 2016-08-03
Allow OG-USA to take general S and J

rickecon:
This PR allows OG-USA to take general S and J. I have tested this on both the steady-state and time path computations for (S=30, J=2) and for (S=80, J=7) and it works perfectly except for the error in the last period resource constraint (1e-03). We think we know what this problem is (nonconvex tax functions) and will remedy it with a different PR.

The steady-state will compute with S=20 and J=1. I assume that fully general S and J would compute if we adjusted our starting values. Most of the changes in this PR were to income.py.


rickecon:
This PR is in reference to [Issue #230](https://github.com/open-source-economics/OG-USA/issues/230). That issue includes some of the discussion and tinkering.

jdebacker:
I'm concerned about the RC error in the last period.  It looks like the tax functions used are concave - the ETR is constant and the MTRx and MTRy are both concave and only depend on x and y, respectively. I might be misremembering, but I thought we got all RC errors to be tiny with this specification.

A check would be to run through with MTRx and MTRy as constants also.  If that yields a last period RC that is within tolerances, then we know it's the tax functions.  Testing with a small S and J should be quick.

rickecon:
@jdebacker @talumbau , Can either of you two see what is failing in the Travis CI checks of this PR? I can't tell by reading the details. I have checked all the usual suspects like the get_reduced_parameters() function in parameters.py. My suspicion is that I am missing the correct TaxFuncEst_baseline.pkl. In other words, if I have 'age specific':True in my kwargs, then I have to have a TaxFuncEst_baseline.pkl that has age-specific tax functions. My suspicion is that the Travis CI call refers to a tax function that is different from what is saved in the repo. But that is just a guess. Any help on getting this PR to pass Travis CI will be appreciated.

rickecon:
@jdebacker I'll run the test with constant (averaged) tax rates.

rickecon:
I just ran the S=30 and J=2 code with the constant tax rates, and the resource constraint error in the last period is now 5.7e-12. I'll run S=80 and J=7 now to make sure that one works.

jdebacker:
@rickecon I see an error about displaying something. Are any of the "show graph" booleans on?

talumbau:
I took a quick look. At least one error is that the code is attempting to create matplotlib figures during a test. The Travis CI environment is not set up to draw these figures so it gets an error. Does it make sense to modify the code so that you can have a flag that turns on drawing the plots but it is off by default (or the other way around and we can modify the tests)?

rickecon:
@talumbau @jdebacker There it is. I had a plotting Boolean left on in the call to inc.get_e_interp() in the parameters.py file (line 689).

rickecon:
@jdebacker @kerkphil , I just finished testing this code with different specifications of the tax functions from txfunc.py and with S=30 and J=2. The only error I ever got was in run #3 in the table below, in which the ETR was a function of both labor and capital income. All the Euler errors and resource constraints in that run were satisfied except for the last period resource constraint error, which was around 1.0e-03.

| Run | ETR form | MTRx form | MTRy form | specific ages | combine ages |
| :---: | --------- | ----------- | ----------- | ------------- | -------------- |
| 1 | const. avg. | const. avg. | const. avg. | good | good |
| 2 | const. avg. | MTRx(x) | MTRy(y) | good | good |
| 3 | ETR(x,y) | MTRx(x) | MTRy(y) | bad | bad |
| 4 | const. avg. | MTRx(x,y) | MTRy(y) | good | good |
| 5 | const. avg. | MTRx(x) | MTRy(x,y) | good | good |
| 6 | const. avg. | MTRx(x,y) | MTRy(x,y) | good | good |

I take this as validation that this PR is ready to be merged and we just need to implement the new functional form for the tax rate functions such that those functions are monotonically increasing in both labor income and capital income, respectively.

PR #242, opened 2016-09-09
Regression test assets

talumbau:


jdebacker:
@TJ, this looks pretty good.  One thing I want to do with tests is to get rid of the `parameters.get_reduced_parameters()`.  It's redundant and can cause falls positive passing of tests.  

However, if we get rid of that option, we might want to add another set of arguments that  is passed to `execute.runner()` and `runnerSS()` in the `kwargs()`.  This would allow one to run a reduced state space model (e.g. `S=40`, `J=2`).  It could be done in two ways:
1) Pass a boolean, `reduced_state`, which would trigger a condition setting of `J`,`lambdas` in `parameters.get_full_parameters()`.  This could be the same as the `QUICK_RUN` boolean you define - the idea would be to pass that through to choose some parameters that help the model run faster.
2) Pass `J` and `lambdas` in the kwargs to `runner()`

I have a preference for option (1).  We don't really need the flexibility of option 2, just something to allow a "full" run and a "fast" run.

jdebacker:
@PeterDSteinberg I see you removed some reforms for temporary testing.  If run time is a strong consideration, I can try to (or advise you) on how we might be able to run a model with a reduced state space that gives us some good indications of changes with lower run times).

@rickecon What do you think about doing some runs with S=40, J=2 for initial regression tests and the ability to kick off runs with S=80, J=7 if things look odd from those tests?

PR #247, opened 2016-10-26
Switched macro output to consider K, not I, thus matching Kss

salimfurth:
For @jdebacker and @rickecon to review: I noticed that the macro_output.py process called by postprocess.py had an inconsistency: the TPI macro columns compared Investment (I) and the SS macro column appended to the end of the same matrix compared Capital (Kss). This minor edit makes them consistent by switching to K for both cases.


jdebacker:
@salimfurth Thanks for noticing the inconsistency.  Let's make them both report the results for Investment (I).  That'll keep all the reported variables as flow variables.

Issue #254, opened 2017-01-17
Fiscal policy baseline


salimfurth:
This issue follows up on #239 
cc'ing @rickecon @jdebacker @MattHJensen 

The fiscal policy baseline currently embodied in OG-USA is made up of numbers picked out of the air that allowed the model to converge. The long-term numbers and transition rules are not a big issue - users shouldn't be reporting results after tG1 (the year when spending is forced up or down to initiate convergence to a user-imposed steady state).

But we do need to decide on a good spending and transfer policy baseline and formalize a clean way for users to implement reforms. There are 3 distinct tasks there, I think:

1. Empirically measure G and T and decide whether the current approach (percent of GDP) is plausible for projections. Some other minor empirical questions fall in here, too.
2. Code those baseline numbers. If we do move away from percent-of-GDP rules of thumb, we will have to think about how to handle the scaling factor.
3. Figure out (or inform those of us who don't know) how to write spending reforms.

This is just my first stab, and I understand the architecture less than Rick & Jason, so feel free to correct me. I'm sure that my employer would be happy to have a paper using this model soon and others may as well.

jdebacker:
@salimfurth thanks for opening this issue.  This is a very useful exercise.

Re 1-2, here's my approach.  I'm using the [OMB Fiscal Year 2017 Historical Tables](https://www.whitehouse.gov/sites/default/files/omb/budget/fy2017/assets/hist.pdf).  
I'll use the following definitions. Transfers = VA Benefits and Services (700) + Income Security (600) + Medicare (570) + Healthcare Services (551). Note that I exclude Social Security from Transfers since we model that component separately in the `theta` parameter.  Government Spending = Total Outlays - Transfers - Net Interest on Debt - Social Security.

I believe that calibrating the percentage-of-gdp values is the way to go to minimize scaling issuing moving between model units and current dollar values (thought with these and our factor adjustment, we should be close - the exception being I don't think there is any way we can get negative real interest rates like we've seen on US gov't debt - debt/gdp ratio and model interest rates won't get us to net interest payments as a share of gdp that we see in 2015 data.

Using 2015 values, I'm finding values for our fiscal parameters of:
```
alpha_T = 0.09
alpha_G = 0.05
initial_debt = 0.59
```

Re 3, I think we might need a call to discuss - I'm not quite sure I know what you are asking.

salimfurth:
[...]

PR #256, opened 2017-01-19
Fix Social Security Calibration

jdebacker:
This PR fixes the calibration of the `theta` parameter, which determines the social security replacement rate.

@rickecon and @kerkphil - can you take a look at the `tax.replacement_rate_vals()` function to see if this approach to calibrating `theta` seems right?

This PR is a work in progress because with the current calibration of `theta`, the model is able to find a solution for the case of per period balanced budgets, but can only solve of the SS (not the TP) when the budget is unbalanced.  I'm not sure of why this is, but the TPI cycles and does not converge.

### DIFF
--- a/Python/ogusa/tax.py
+++ b/Python/ogusa/tax.py
@@ -35,41 +35,29 @@ def replacement_rate_vals(nssmat, wss, factor_ss, params):
         theta      = [J,] vector, replacement rates by lifetime income group
     Returns: theta
     '''
-    e, J, omega_SS, lambdas = params
-
-    # Do a try/except, depending on whether the arrays are 1 or 2 dimensional
-    try:
-        AIME = ((wss * factor_ss * e * nssmat) *
-                omega_SS).sum(0) * lambdas / 12.0
-        PIA = np.zeros(J)
-        # Bins from data for each level of replacement
-        for j in xrange(J):
-            if AIME[j] < 749.0:
-                PIA[j] = .9 * AIME[j]
-            elif AIME[j] < 4517.0:
-                PIA[j] = 674.1 + .32 * (AIME[j] - 749.0)
-            else:
-                PIA[j] = 1879.86 + .15 * (AIME[j] - 4517.0)
-        theta = PIA * (e * nssmat).mean(0) / AIME
-        # Set the maximum replacment rate to be $30,000
-        maxpayment = 30000.0 / (factor_ss * wss)
-        theta[theta > maxpayment] = maxpayment
-    except:
-        AIME = ((wss * factor_ss * e * nssmat) *
-                omega_SS).sum() * lambdas / 12.0
-        PIA = 0
-        if AIME < 749.0:
-            PIA = .9 * AIME
-        elif AIME < 4517.0:
-            PIA = 674.1 + .32 * (AIME - 749.0)
+    e, S, J, omega_SS, lambdas, retire = params
+    dim2 = 1
+    if nssmat.ndim==2:
+        dim2 = J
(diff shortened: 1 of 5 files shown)


kerkphil:
@jdebacker , @rickecon - the code looks good to me.  I'm trying to remember what the replacement rate is conceptually in the model.  From the code we have: PIA = theta * AIME / LifetimeEffectiveLabor.  Why do we need the lifetime effective labor adjustment?

jdebacker:
@kerkphil I think the main change in the way `theta` is calculated now is that the AIME is based on the last 35 years of earnings.  But there is also a change in how the PIA `theta` are computed that better follows my understanding of the law.  In particular, given that the PIA is the insurance amount, this line of code:

`theta = PIA * (e * nssmat).mean(0) / AIME`

Didn't make sense to me.   Currently we are doing:

`theta = (PIA*(12.0*S/80)) / (factor_ss*wss)`

Which taxes the PIA, turns it from a monthly amount to an annual amount.  It then transforms it from dollars to model units via the division by the `factor`.  The wage rate in the denominator is because benefits adjusted by wage growth, so we scale by wages.  When we put benefits into the budget of model households, we do `theta*w`.

kerkphil:
@jdebacker OK, I looked at the old code.  Let me look at the PR code, but your explanation of theta makes a lot more sense to me.

kerkphil:
@jdebacker Yes, this code looks better to me.  One issue, is AIME based on the past 35 years of wages or the 35*12 highest months of adjusted wages over the workers whole lifetime?  My impression was the latter.

jdebacker:
@kerkphil - yes, you are right that it's the highest 35 years.  Thinking more about it - that's something we should be able to do.  Let me update the PR to reflect that.

kerkphil:
@jdebacker Since we are looking at the steady state values, the highest 35 years of wages are probably the 35 years immediately before retirement anyway, so it may not make that much difference.  However, if it's not too difficult it would be best to recode, just to be safe.

Python/ogusa/tax.py
```diff
@@ -35,41 +35,27 @@ def replacement_rate_vals(nssmat, wss, factor_ss, params):
...
-            PIA = 1879.86 + .15 * (AIME - 4517.0)
-        theta = PIA * (e * nssmat).mean(0) / AIME
-        # Set the maximum replacment rate to be $30,000
-        maxpayment = 30000.0 / (factor_ss * wss)
-        if theta > maxpayment:
-            theta = maxpayment
-    theta = 0  # setting theta = 0 since we are including social security income in capital income (CHECK)
+            PIA[j] = 1879.86 + .15 * (AIME[j] - 4517.0)
+    # Set the maximum replacment rate to be $30,000
+    maxpayment = 3501.00 #30000.0/12.0
```
jdebacker: @kerkphil I updated the AIME calculation to take the highest 35 years of earnings.  I also changed the PIA max amount.  $30,000 was an annual amount, I think.  Replaced with $3501 per month.  Not 100% I got this right.  Took highest number from 2015 from [this table](https://www.ssa.gov/oact/cola/examplemax.html).

kerkphil: Looks great!

jdebacker:
TPI not seems to converge (not cycling).  Two updates helped with this:
1) Better initial guesses
2) A smoother transition path for government spending.

jdebacker:
Fiscal parameters are now updated per discussion in Issue #254.  

Removing WIP.

jdebacker:
Merging.  Will address further fiscal calibration issues in another PR.

PR #257, opened 2017-01-26
Corporate tax

jdebacker:
This PR adds a corporate income tax on the representative firm.  This allows for some level of business tax analysis in OG-USA and also helps with the calibration of the fiscal parameters.

A future PR will add differentiation of the firms (e.g. corporate and non-corporate, etc).


salimfurth:
Not technically related to these changes, but the previous PR introduced an error when running with small_open=True and budget_balance=False in Line 713 of TPI.py. It's an easy fix - D & G just need to be created earlier. We could include it in this PR or do a separate one.

jdebacker:
@salimfurth - thanks for pointing that out.  I'll take a look and fix in this PR.

salimfurth:
A more substantive question, @jdebacker: would it be easy to put in a deduction, so that the marginal rate > average rate? That would go a long way toward matching the incentives faced by real world firms.

jdebacker:
Yes - we can do that.  Let me open an issue to discuss.

PR #258, opened 2017-01-26
regression tests fixes to take .regression.txt budget_balance bool

PeterDSteinberg:
Some minor fixes to the regression testing system for Jenkins related to taking the `budget_balance` boolean argument from `.regression.txt` and ensuring it is passed to `execute.py`'s `runner` function.

Also some changes in `polling_jobs.py` that are only relevant to Jenkins testing (how errors and logging are handled regarding jobs whose builds cannot be found, as in the case of a job queued for 12 hours before starting).


PeterDSteinberg:
Note - leave this PR open as it helps with regression testing of OG-USA when we need to do that.

PR #267, opened 2017-02-28
Remove passing growth assumptions to taxcalc

jdebacker:
This PR removes the passing of growth assumptions defined by the OG-USA user to the Tax Calculator.  This makes OG-USA compatible with TaxCalc 0.7.7.

Note that this passing of the exogenous long run growth rate from OG-USA to Tax Calc was not the correct mapping between the growth parameters of the two models.  If one wants to correctly pass growth rates from OG-USA to Tax Calc, one would need to first solve the OG-USA model to recover the endogenous growth rates over the budget window period.  Further, to be consistent, one would need to iterate between the Tax Calc and OG-USA to ensure that the growth rates are consistent with the tax functions derive from Tax Calc with those growth assumptions.

### DIFF
--- a/Python/ogusa/get_micro_data.py
+++ b/Python/ogusa/get_micro_data.py
@@ -28,37 +28,6 @@ import copy
 import numba
 import pickle
 
-def only_growth_assumptions(user_mods, start_year):
-    """
-    Extract any reform parameters that are pertinent to growth
-    assumptions
-    """
-    growth_dd = taxcalc.growth.Growth.default_data(start_year=start_year)
-    ga = {}
-    for year, reforms in user_mods.items():
-        overlap = set(growth_dd.keys()) & set(reforms.keys())
-        if overlap:
-            ga[year] = {param:reforms[param] for param in overlap}
-    return ga
-
-
-def only_reform_mods(user_mods, start_year):
-    """
-    Extract parameters that are just for policy reforms
-    """
-    pol_refs = {}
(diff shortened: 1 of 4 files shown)


PeterDSteinberg:
I'll test this out now.

jdebacker:
@PeterDSteinberg Let me know what you find testing this.  Travis CI is failing, but I'm not seeing any log when I click "Details", so I'm not sure what is causing the failure.

jdebacker:
@PeterDSteinberg Ok, log is now there.  Maybe yesterday's issue related to Amazon's problems?

In any case, seeing log now, appears to be issue with Tax Calcuator call, but I don't know fix. Ideas?

PeterDSteinberg:
@jdebacker Issues yesterday (Feb 28) would be expected given S3 troubles and the fact that heroku depends on AWS.

PeterDSteinberg:
One thing @jdebacker is that this OG-USA CI testing uses the taxcalc conda package.  If you have tested out locally using Tax-Calculator master then that may explain some issues.  I need to make conda packages for all repos, but was waiting for this PR to get finalized.   I go back and forth on the best approach for CI in situations like this with multiple repos: should it install the bleeding edge case master or the latest package?  Try it out locally with Tax-Calc master.  If that fixes the problems let me know and I'll build Tax-Calc now so the CI hopefully passes too.  I have to put some time into another project now and will check back in on this Friday afternoon.

jdebacker:
@PeterDSteinberg Will do. Thanks.

PR #278, opened 2017-03-22
[WIP] Generalize the firm production function

jdebacker:
This PR is the first step in adding representative firms from multiple industries into the model.  In particular, this PR generalizes the firm production function from a Cobb-Douglas function to a more general Constant Elasticity of Substitution production function that nests the CD function as a special case.

The [WIP] label will be left on here until extensive testing is completed.


salimfurth:
This is tremendously valuable. Let me know what kind of help/testing would be helpful, @jdebacker.

jdebacker:
In addition to extending to a general CES function, this PR fixed a few outstanding issues:
1) Issue #284: There was a need to add a second function to solve the reform policies in the case `baseline_spending=True`.  One function could not handle both options.
2) Issue #283: The value for the GDP variable, `Y`, was not updating at each iteration of the TPI solution method.
3) There was a bug in the determination of the social security replacement rate.  It had been based on earnings only in the baseline SS.  This PR ensures that the replacement rate value is determined by steady state earnings in the relevant steady state (i.e., the values differ between the baseline and reform).

I tested this method for all ten combinations of `small_open`, `budget_balance`, and `baseline_spending` that are possible, for both the baseline and a single policy reform.  The SS and the TPI in test mode, all solved fine (i.e. all criteria met the required tolerances) 

I ran TPI with `test=False` (i.e., the full model) for the baseline and reform policies with `small_open=True`, `budget_balance=False`, and `baseline_spending=False` and found the model to solve fine.

Running the TPI with `test=False` (i.e., the full model) for the baseline and reform policies with `small_open=False`, `budget_balance=False`, and `baseline_spending=False` and found the model to solve fine for the baseline, but that a discrete jump of 2e-6 appears in the final period resource constraint.  Thus the model returned and errors as the default tolerance of 1e-9 for the resource constraint error was not met.  This was with an elasticity of substitution of capital and labor = 0.65.  When setting this elasticity to 1 (i.e., going back to Cobb-Douglas), there was not such problem.  We should keep an eye on this, but the issue could have been a result of the percentage deviation distance measure used in the TPI and the fact that the SS interest rate with an elasticity = 0.65 was negative and very close to zero - thus small absolute differences resulted in large percentage changes.

PR #305, opened 2017-07-07
Change travis package pins

jdebacker:
This PR will change the `travis.yml` file to pin the packages to versions that work with OG-USA.


jdebacker:
@PeterDSteinberg @rickecon I think I fixed the issues with the environment specification both on Travis CI and for the local environment.  

Could one of you review and merge if you can run the program without errors?  BTW, you may want to set the `test` boolean to `True` if you test with `run_ogusa_serial.py` and want to go through the entire time path.

Issue #290, opened 2017-04-05
2 setup.py issues


PeterDSteinberg:
* [ ] Make OG-USA uninstall cleanly so that updating from a source install or conda package does not require `conda remove` or `pip uninstall`.  If tricky, then document these commands as update steps.
 * [ ] It's my impression that `python setup.py develop` does not work the way `python setup.py install`.  This is a reminder issue to follow up on that and fix if that is the case.

hdoupe:
@PeterDSteinberg 

> It's my impression that python setup.py develop does not work the way python setup.py install. This is a reminder issue to follow up on that and fix if that is the case.

Here is the error:

```
(ogusatemp) HDoupe-MacBook-Pro:OG-USA henrydoupe$ python setup.py install
running install
running bdist_egg
running egg_info
writing ogusa.egg-info/PKG-INFO
writing top-level names to ogusa.egg-info/top_level.txt
writing dependency_links to ogusa.egg-info/dependency_links.txt
reading manifest file 'ogusa.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
writing manifest file 'ogusa.egg-info/SOURCES.txt'
installing library code to build/bdist.macosx-10.7-x86_64/egg
running install_lib
running build_py
UPDATING build/lib/ogusa/_version.py
set build/lib/ogusa/_version.py to '0.5.8+6.ge357548.dirty'
creating build/bdist.macosx-10.7-x86_64/egg
creating build/bdist.macosx-10.7-x86_64/egg/ogusa
copying build/lib/ogusa/__init__.py -> build/bdist.macosx-10.7-x86_64/egg/ogusa
copying build/lib/ogusa/_version.py -> build/bdist.macosx-10.7-x86_64/egg/ogusa
copying build/lib/ogusa/calibrate.py -> build/bdist.macosx-10.7-x86_64/egg/ogusa
creating build/bdist.macosx-10.7-x86_64/egg/ogusa/data
creating build/bdist.macosx-10.7-x86_64/egg/ogusa/data/ability
copying build/lib/ogusa/data/ability/FR_wage_profile_tables.xlsx -> build/bdist.macosx-10.7-x86_64/egg/ogusa/data/ability
creating build/bdist.macosx-10.7-x86_64/egg/ogusa/data/demographic
copying build/lib/ogusa/data/demographic/demographic_data.csv -> build/bdist.macosx-10.7-x86_64/egg/ogusa/data/demographic
[...]

PR #308, opened 2017-07-17
Restructure directories

hdoupe:
This PR addresses issue #290.  The `Python` directory was removed and its contents were moved into the `OG-USA` directory.  The `run_*.py` scripts were placed in the `run_examples` directory.

This was done with the following commands
```
cd OG-USA
mv Python/* .
rm -r Python/
mv run_*.py run_examples/
```

The paths in several files were modified so that they did not reference the `Python` directory.  Now you can run `python setup.py develop`, `cd ogusa`, and py.test works.


jdebacker:
@hdoupe Thanks for making this PR.  It'd be nice to pass Travis CI.  I looks like the failure might have to do with the change in directory structure and the test not being able to find some files.  I noticed you changed a path in the `.travis.yml` script, which looks good.  But it there an issue with an OG-USA package?  

cc @PeterDSteinberg

PeterDSteinberg:
Looks like it can't find this pickle file: `TxFuncEst_policy.pkl`  @jdebacker Can you give me some background on that pickle file, like
 * Is it created on the fly ideally (like you run OG-USA once and then it exists and can be used opportunistically to make it faster next time)? 
 * Is it something from one of the repos? 
 * Does it have public use file details in it or anything we need to worry about?
 * Does it always get created in a specific directory relative to a code file or the current directory or look for an environment variable to guess where it is?

I'll make a separate issue to figure out a Python version-independent serialization strategy to be used in place of pickle (as we discussed in the meeting) and may mention some of your answer on that new issue.

jdebacker:
@PeterDSteinberg asks:
> Is it created on the fly ideally (like you run OG-USA once and then it exists and can be used opportunistically to make it faster next time)?

Files like these are created on the fly for evaluating policy.  But two pre-estimated sets of parameters are stored in the repo for testing purposes (`TxFuncEst_baseline.pkl` and `TxFuncEst_policy.pkl`).  `TxFuncEst_baseline.pkl` and `TxFuncEst_policy.pkl` reside in the repository for testing the model without need to run the `get_micro_data.py` and `txfunc.py` scripts with take 2+ hours of run time.  These scripts estimate the parameters of tax functions that are used by the OG-USA model.  

When OG-USA is run to evaluate policy (whether locally or on TaxBrain) then pickle files with the same stored objects are created.  The two files created correspond to the baseline and reform policies.  These are saved with a `guid`  in the filename to help know what run these parameters correspond to.  

> Is it something from one of the repos?

Yes, 2 pickles with the tax function parameters are kept in the repo for testing purposes.  See above.

> Does it have public use file details in it or anything we need to worry about?

No.  These parameters are estimated using the micro data, but do not have any detail on that data and each parameter summarizes at least 600 observations.

> Does it always get created in a specific directory relative to a code file or the current directory or look for an environment variable to guess where it is?

Yes.  The path is assigned [here](https://github.com/open-source-economics/OG-USA/blob/master/Python/ogusa/parameters.py#L46) in `parameters.py` and [here](https://github.com/open-source-economics/OG-USA/blob/master/Python/ogusa/txfunc.py#L48) in `txfunc.py`.

hdoupe:
@jdebacker The tests pass on my machine.  My understanding is that these files should be in the directory above `ogusa`.  So, in the new structure, they would be in the `OG-USA` directory and in the old structure, they would be in the `Python` directory.

It appears that the Travis CI test is looking for these files in the correct place but they aren't there
```
E       IOError: [Errno 2] No such file or directory: '/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/ogusa-0.5.8+9.g87cff93-py2.7.egg/TxFuncEst_policy.pkl'
```

@PeterDSteinberg

hdoupe:
The solution was to add a path for `TxFuncEst_policy.pkl` to the `package_data` parameter in `setup.py`.

@jdebacker @PeterDSteinberg

jdebacker:
@hdoupe Thanks for this contribution.  These changes will be really helpful on a number of fronts. Merging.

PR #311, opened 2017-07-19
Remove ancillary files

jdebacker:
This PR removes a number of ancillary directories and files per the discussion in Issue #290.  At some point, these data and documentation files may be added to another repository, but we want to clean up the OG-USA repo so that is contains only the files necessary to run and test the model.


jdebacker:
@rickecon Let me know what you think of this.  I've saved all files removed from this repo so that we can add them somewhere else at some point.

PR #313, opened 2017-08-16
Add aggregates module

jdebacker:
This PR adds a modules called `aggregates.py` that defined functions used to compute economic aggregates. All these functions were defined before, but contained in `firm.py` or `household.py`.  It is now the case that `firm.py` and `household.py` contain only functions necessary to solve an individual household or firm problem and any economic aggregates are computed in `aggregates.py`

A new unit test, `test_aggregates.py` was added to include unit testing of the functions that used to be testing in `test_firm.py` but that were aggregate functions.


rickecon:
@jdebacker . This looks ready to me. Are you ready for me to merge this?

jdebacker:
@rickecon, yes, this is ready to merge.

PR #317, opened 2017-10-12
changes for world interest rate - small_open economy

PeterDSteinberg:
This the OG-USA related change associated with [Policybrain issue #481](https://github.com/OpenSourcePolicyCenter/PolicyBrain/issues/481).  See @rickecon 's comment near the bottom of that issue where he talks about the OG-USA parameters to pass world interest rate (that affects several other parameters' default choices).

### DIFF
--- a/ogusa/scripts/execute.py
+++ b/ogusa/scripts/execute.py
@@ -49,7 +49,7 @@ def runner(output_base, baseline_dir, test=False, time_path=True, baseline=False
         txfunc.get_tax_func_estimate(baseline=baseline, analytical_mtrs=analytical_mtrs, age_specific=age_specific,
                                      start_year=user_params['start_year'], reform=reform, guid=guid)
     print 'In runner, baseline is ', baseline
-    if small_open and (not isinstance(small_open, dict) or not all(key in small_open for key in SMALL_OPEN_KEYS)):
+    if small_open and (not isinstance(small_open, dict)):
         raise ValueError('small_open must be False/None or a dict with keys: {}'.format(SMALL_OPEN_KEYS))
     small_open = small_open or {}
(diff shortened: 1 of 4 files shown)


PeterDSteinberg:
@rickecon @jdebacker - Let me know if this seems reasonable.  (I'm still working on corresponding django app changes)

ogusa/scripts/execute.py
```diff
@@ -47,10 +49,15 @@ def runner(output_base, baseline_dir, test=False, time_path=True, baseline=False
         txfunc.get_tax_func_estimate(baseline=baseline, analytical_mtrs=analytical_mtrs, age_specific=age_specific,
                                      start_year=user_params['start_year'], reform=reform, guid=guid)
     print 'In runner, baseline is ', baseline
-    run_params = ogusa.parameters.get_parameters(test=test, baseline=baseline, guid=guid)
+    if small_open and (not isinstance(small_open, dict) or not all(key in small_open for key in SMALL_OPEN_KEYS)):
```
jdebacker: @PeterDSteinberg You changes generally look good.  These lines of code have a lot going on and make things a bit opaque.  They look more complicated than what I thought would be the case for just adding one more user modifiable parameter, but I see that you are trying to have the boolean for "small open" and the world interest rate parameter together in just one argument and if that's the case I don't have an immediate suggestion of a cleaner way to do this.

PeterDSteinberg: @jdebacker I simplified that part a bit, but it stands out mainly because it is the only parameter that is validated.

jdebacker:
@PeterDSteinberg If you can resolve the conflicts here, this PR is good to go.  Thanks!

jdebacker:
Replaced by PR #335

PR #319, opened 2017-10-25
add warning about budget balance < 0.

PeterDSteinberg:
Part of the work on [Policybrain issue 482](https://github.com/OpenSourcePolicyCenter/PolicyBrain/issues/482)


hdoupe:
@PeterDSteinberg How do you want to relay this message back to TaxBrain?  For file uploads, a list of error/warning messages is sent to the file upload page in the context.  This list is simply displayed at the top of the page.  I think we could handle this warning message in a similar way, but the list of warning messages would be sent along with the results to the results page.

What do you think?

PeterDSteinberg:
@hdoupe - The front end in Policybrain will need to check for `Gss`'s value and make the warning there - you're right this PR only adds a warning message.  We could also add a message field, but I think it is better in general for the front end to design its own error/warning message wording and processing to better separate docs and front end from backend.

jdebacker:
@PeterDSteinberg if you can resolve the conflict here, then this is good to go.  Thanks!

jdebacker:
Replaced by PR #334

PR #320, opened 2017-10-25
Updates to open econ

jdebacker:
This PR updates the SS solution algorithm for the open economy version of the model.

This fixes a bug where the wage did not update in a way that was consistent with the firm's FOC.

NOTE: this PR does not entirely fix the the problem.  At a minimum, one needs to extend these fixes to the time path solution and write the `firm.get_w_from_r()` function in a way that allows for general CES production functions (currently it is only written for Cobb-Douglas production functions).

It is not necessary, but with this method, one could also dispense with the guess of `w` in the GE loop of the solution.  For clarity of the model, that change should be added as well sometime soon.


rickecon:
Looks great @jdebacker. Let's not merge this PR until we add the corresponding fix to `TPI.py`.

jdebacker:
@rickecon Review this, but I think this is ready to go.  Combined with PR #325, this should make it substantially easier to run the reforms you have been trying recently and to compare open vs closed economy results.

The only thing left from here is to update `firm.get_w_from_r()` for the case of a non-Cobb-Douglas production function.  But so long as you run with CD, results should be accurate.

PR #322, opened 2017-10-27
Update OG-USA environment

jdebacker:
This PR updates the OG-USA environment to be consistent with the environment used for other PolicyBrain tools.

A few adjustments were made to the `demographics.py` script to work with the updated version of NumPy.

The SS works fine.  I am still running the full tax function estimation and time path equilibrium.  Hold off on merging until those are reviewed.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -5,12 +5,12 @@ dependencies:
 - ipython
 - ipython-notebook
 - setuptools
-- nomkl
-- scipy=0.18.1
+- mkl
+- scipy>=0.18.1
 - matplotlib
-- taxcalc=0.8.3
+- taxcalc>=0.12.0
 - numba
-- numpy=1.11.2
-- pandas==0.18.0
+- numpy>=1.12.1
+- pandas>=0.20.1
 - pytest
 - python=2.7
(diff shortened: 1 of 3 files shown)


jdebacker:
Update: The time path solution completed and results look very similar to what we had before (differences from prior runs appear small enough that I think they are due to updating from Tax-Calc 0.8.3 to 0.12.0).

If this satisfied the environment for PolicyBrain, I will merge.  @hdoupe let me know.

hdoupe:
This is great.  Thanks again for the quick response.  This should be fine.  I looked through the changes you made and they appear pretty minor.  Which reform did you use to run the model?

I added a quick [test](https://github.com/hdoupe/OG-USA/blob/138e04c81af6707ba355aa8f9fa35df5c03f7f3e/regression/regression_tests.py#L28) to compare the macro output data in PR #316 .  If you used a reform that was used in the regression tests, we could compare the results and be a little more certain that nothing broke in the update.

jdebacker:
@hdoupe I just ran the baseline.  I did get the exact same results using old tax functions, so I think we are ok.   In my run over night, the tax functions would have changed (slightly) due to changes in tax-calculator.

I'll go ahead and merge.

hdoupe:
That makes sense. Thanks

rickecon:
I see @talumbau adding and removing "in progress" labels to these issues. Much appreciated. I have to believe that TJ has set up a GitHub bot on his account to do this. Otherwise, TJ is a very vigilant issue label maintainer.

andersonfrailey:
>I see @talumbau adding and removing "in progress" labels to these issues. Much appreciated. I have to believe that TJ has set up a GitHub bot on his account to do this. Otherwise, TJ is a very vigilant issue label maintainer.

Personally, I like the believe it's the latter.

PR #323, opened 2017-11-03
Set path to tax function output

hdoupe:
This PR allows the user to set the path for the tax function estimation output.  A test was added in `test_parameters.py` to test the logic in `parameters.get_parameters`.  

I tried to write a test, `test_file.py`, that checked that output files were written to the correct location.  However, this involved mocking many functions, and this messed up the other tests.  Unless we find a better way to test the file writing behavior, this should be removed from the PR.


hdoupe:
@jdebacker This PR is ready for review.

jdebacker:
@hdoupe This looks good to me.  Will merge soon - wanting to wait until @rickecon gets some runs done that he's in the middle of first.

hdoupe:
Thanks @jdebacker.  Contingent on passing the the travis tests, this should be good to go.

jdebacker:
@hdoupe these and some other PRs are looking good and I'm on board to merge.  Hesitating because @rickecon  is in the middle of some runs and I think he'll have some changes.

@rickecon Let me know when you are ok with merging these and we can go ahead and do that.  Let's merge these before you add your PRs for fixes you uncovered trying to run TCJA.

hdoupe:
Ok, that sounds good.  The merge order should probably be #323, #327, and then #324.  I think #324 will have some merge conflicts that will need to be resolved.

PR #324, opened 2017-11-03
Allow user to specify data argument to Tax-Calculator

hdoupe:
This PR allows the user to specify a data object to be passed to `taxcalc.Records` as the `data` keyword argument.   This allows the user to specify whether they want to use `puf.csv`, `puf.csv.gz`, or an already loaded pandas dataframe.  Further, it would be easy to add behavior such as setting `data` equal to "cps" implies that the `taxcalc.Records.cps_constructor` method should be used.


jdebacker:
@hdoupe Merged your other two PRs.  Please resolve the conflicts here and then this one will be good to go too.  Thanks!

hdoupe:
@jdebacker @rickecon I updated `get_micro_data.get_data()` to increment to year 2027.  Are there any other changes that need to be made to correspond to the new Tax-Calculator data?

ogusa/get_micro_data.py
```diff
@@ -66,13 +69,16 @@ def get_calculator(baseline, calculator_start_year, reform=None, data=None, weig
...
         calc1.increment_year()
 
+    # running all the functions and calculates taxes
+    calc1.calc_all()
+
     return calc1
 
 
-def get_data(baseline=False, start_year=2016, reform={}):
+def get_data(baseline=False, start_year=2016, reform={}, data=None):
```
jdebacker: We might make the default `start_year` be 2018 at this point.

hdoupe: Thanks, this is fixed in the latest commits.

ogusa/parameters.py
```diff
@@ -301,7 +301,6 @@ def get_parameters(test=False, baseline=False, guid='', user_modifiable=False, m
     T = int(4 * S)
     BW = int(10)
 
-    start_year = 2016
     starting_age = 20
```
hdoupe: @jdebacker Will setting this to 2018 break anything?  Is there a reason why this was hard coded?

jdebacker: @hdoupe This needs to be fixed.  I think hard coding comes from very early on before we were passing a `start_year` variable.  I don't think changing this to 2018 will be an issue.  

*But* what we need to do have start_year set only in one place.  Besides this hard coded value, it is currently in the `user_params` dictionary (specified in the runner script and passed into `execute.py`) but the value in `parameters.py` is not overwritten with this user defined value.  The value in the `user_params` dictionary only affects the tax functions estimation (see line 50 of `execute.py`).  We want it to also affect the demographics in the call to `dem.get_pop_objs()` in `parameters.py`.

I think the best way to do this is to make `start_year` a kwarg for `parameters.get_parameters()`.  But I'm open to other methods.

ogusa/parameters.py
```diff
@@ -137,7 +137,7 @@ def get_parameters_from_file():
 
 
 def get_parameters(test=False, baseline=False, guid='', user_modifiable=False, metadata=False,
-                   tx_func_est_path=None):
+                   tx_func_est_path=None, start_year=2018):
     '''
```
hdoupe: @jdebacker `start_year` is added as a keyword argument where `get_parameters` is called in `execute.py`.  Is this what you meant?

ogusa/scripts/execute.py
```diff
@@ -61,7 +61,8 @@ def runner(output_base, baseline_dir, test=False, time_path=True,
         test=test, baseline=baseline, guid=guid,
         tx_func_est_path=os.path.join(
             output_base,'TxFuncEst_{}.pkl'.format(guid)
-        )
+        ),
+        start_year=user_params['start_year']
     )
```
hdoupe: And here `start_year` is passed as a key word argument to `get_parameters`.

ogusa/parameters.py
jdebacker: @hdoupe Yes - so line 59 of `execute.py` would be:

```
run_params = ogusa.parameters.get_parameters(
        test=test, start_year=start_year=user_params['start_year'], baseline=baseline, guid=guid,
        tx_func_est_path=os.path.join(
            output_base,'TxFuncEst_{}.pkl'.format(guid)
```

ogusa/scripts/execute.py
jdebacker: Not sure what you are asking here - I think how `start_year` is passed to the tax function estimation is fine here and am suggesting we do similar to pass it to `parameters.get_parameters()`.

hdoupe: Right, but this is where `start_year` is passed to `parameters.get_parameters()`.

jdebacker: Ok - yes, this is the change we need.  Thanks!

hdoupe: Great, no problem.

hdoupe:
@jdebacker This should be ready now.

rickecon:
@hdoupe @jdebacker . Changing the `start_year` to 2018 doesn't break any of the code that I have been running in OG-USA. I think this is ready to go.

jdebacker:
Thanks @hdoupe and @rickecon, merging now.

PR #334, opened 2017-12-12
Add Gss warning

jdebacker:
This PR replaces PR #319 and adds a warning message if government spending is forced negative to balance the budget in the steady-state.  The model solves, but the user is notified of this condition.

### DIFF
--- a/ogusa/SS.py
+++ b/ogusa/SS.py
@@ -28,6 +28,7 @@ from . import aggregates as aggr
 import firm
 import utils
 import os
+import warnings
 
 
 '''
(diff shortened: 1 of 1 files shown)


hdoupe:
@jdebacker do you think it would make sense to add a capture warnings option in `execute.runner`?  

I'm thinking about this in the context of capturing the warnings from PolicyBrain and displaying them above the output.  It seems like this PR only writes them to stdout.  There's a [`capture_warnings` function in the logging module](https://docs.python.org/2/library/logging.html#logging.captureWarnings).  So that could be a good place to start.

jdebacker:
@hdoupe I would like the warning to display above the output in PolicyBrain.  I know nothing about `capture_warnings`, but will look into it. Thanks for the suggestion.

jdebacker:
@hdoupe Have you use the `capture_warnings` function before?  While it seems straightforward to log the warnings, I'm wondering where this should be placed so that PolicyBrain could get that info.  Could you direct me on that?  Glad to get on a  call tomorrow if that's easier.

PR #339, opened 2018-01-07
fix if statments in TPI

jdebacker:
This PR makes the `if` statements in `TPI.py` more Pythonic, avoiding some unexpected results when running the open economy version of OG-USA.

### DIFF
--- a/ogusa/SS.py
+++ b/ogusa/SS.py
@@ -589,7 +589,7 @@ def SS_solver(b_guess_init, n_guess_init, rss, T_Hss, factor_ss, Yss,
         if budget_balance:
             debt_ss = 0.0
         else:
-            debt_ss = debt_ratio_ss*Y
+            debt_ss = debt_ratio_ss * Y
         Kss = Bss - debt_ss
         Iss_params = (delta, g_y, omega_SS, lambdas, imm_rates,
                       g_n_ss, 'SS')
--- a/ogusa/SS.py
+++ b/ogusa/SS.py
@@ -356,7 +356,7 @@ def inner_loop(outer_loop_vars, params, baseline, baseline_spending=False):
                        full_output=True)
 
         euler_errors[:, j] = infodict['fvec']
-      # print 'Max Euler errors: ', np.absolute(euler_errors[:,j]).max()
+        # print('Max Euler errors: ', np.absolute(euler_errors[:,j]).max())
 
         bssmat[:, j] = solutions[:S]
         nssmat[:, j] = solutions[S:]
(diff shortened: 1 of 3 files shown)


rickecon:
@jdebacker . Is this ready to go? That is, have you tested it with the options that the "if" statements affect?

jdebacker:
@rickecon I haven't.  It shouldn't be any worse than before.  But I should write a script to run through the six possible configurations of the model.  I'll do that today and report back.

hdoupe:
@jdebacker do you think you could write some tests that fail using the code before you made the changes and pass with the new changes?

jdebacker:
@hdoupe Sure.  I may need to consult you on how best to structure, though.

hdoupe:
No problem.  I would just use the normal `sim_params` dictionary that's passed to `create_TPI_params`, alter the key-value pairs for `baseline` and `baseline_spending`, and check to make sure that you get the expected results.

Testing this in the `run_TPI` function looks pretty difficult and doesn't seem like it's worth the effort.  However, I think this may be a good time to start thinking about ways to get some of the larger functions like `run_TPI` under test so that they can be refactored safely if that's something that you and @rickecon are interested in.

jdebacker:
@rickecon Here's what I have seen so far when trying to test different version of the model with this update:

1) Open economy, balanced budget runs as expected.
2) Open economy, unbalanced budget, baseline spending=False runs as expected.
3) Open economy, unbalanced budget, baseline spending=True runs as expected.
4) Closed economy, unbalanced budget, baseline spending=False runs as expected *only for full scale model*.  i.e., `test=True` where smaller state spaced is used does not solve.
5) Closed economy, unbalanced budget, baseline spending=True runs as expected *only for full scale model*.  i.e., `test=True` where smaller state spaced is used does not solve.
6) Closed economy, balanced budget *does not run as expected* - there is a RC error in TPI that raises an exception.

I am going to look into (6) first.  I am not sure what to do about (4) and (5).  Being able to run in "test mode" is very helpful to speed up tests.  It is also odd that model is having issues there (though I'm not so worried because results seem fine with the full state space).

jdebacker:
With most recent commit, (6) is solved.  Issue had to do with values for debt over the time path in the balanced budget case.  May have been caused as a result of changes way back in PR #255 when the user defined SS debt-to-GDP ratio was created.

I'm still unsure why the version of the model in (4) and (5) do not run with the reduced state space.  But after a few hours I didn't see anything wrong here - just that the root finder does not work well in this case and starts throwing `nan`'s.

PR #341, opened 2018-01-17
Investment in output pickles

jdebacker:
This PR adds total investment (not just domestic investment) and total domestic savings to the output dictionaries for the SS and TPI output.

### DIFF
--- a/ogusa/SS.py
+++ b/ogusa/SS.py
@@ -693,6 +693,9 @@ def SS_solver(b_guess_init, n_guess_init, rss, T_Hss, factor_ss, Yss,
     # np.savetxt("mtr_ss_capital.csv", mtry_ss, delimiter=",")
     # np.savetxt("mtr_ss_labor.csv", mtrx_ss, delimiter=",")
 
+    # Compute total investment (not just domestic)
+    Iss_total = delta * Kss
+
     # solve resource constraint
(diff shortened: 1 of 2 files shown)


jdebacker:
@rickecon I just added a commit to include corp and individual revenue separately in the SS and TPI output pickles.  This will be helpful as we make sure we hit revenue targets.

jdebacker:
@hdoupe Did something recently change with CPS file that would cause the `test_cps` test to fail?  It looks like a simple test. Could here be an issue with the start year?

hdoupe:
Tax-Calculator 0.15.0 changed how a lot of the attributes are accessed, including the `Records` object and its attributes. I think you would re-write that line as:

```
assert (calc.array("blind_head").sum() > 0 and
                calc.array("e00700").sum() == 0)
```

You may want to do a separate PR to upgrade to TC 0.15.0. This would be required to get back on PolicyBrain. Hopefully, in the future we will be able to support multiple environments, but for now, everyone is stuck in the same environment. Sorry, I should have mentioned this on the call this morning.

jdebacker:
@hdoupe OK - thanks for the heads up.  I'll make that change in a separate PR.  

@rickecon If you want to merge this now to help with splitting out revenue, then feel free.  This branch is running fine locally with TC 0.14.3.

PR #346, opened 2018-01-26
Test for different start years

jdebacker:
This PR adds a test to ensure that OG-USA runs with any start year from 2016 to 2026.

### DIFF
--- a/ogusa/parameters.py
+++ b/ogusa/parameters.py
@@ -371,9 +371,14 @@ def get_parameters(output_base, reform={}, test=False, baseline=False,
         raise RuntimeError(err)
 
     # Income Tax Parameters
-    tx_func_est_path = os.path.join(
-        output_base, 'TxFuncEst_{}.pkl'.format(guid),
-    )
+    if baseline:
+        tx_func_est_path = os.path.join(
+            output_base, 'TxFuncEst_baseline{}.pkl'.format(guid),
+        )
+    else:
+        tx_func_est_path = os.path.join(
+            output_base, 'TxFuncEst_policy{}.pkl'.format(guid),
+        )
     if run_micro:
         txfunc.get_tax_func_estimate(BW, S, starting_age, ending_age,
                                      baseline=baseline,
--- a/ogusa/parameters.py
+++ b/ogusa/parameters.py
@@ -374,7 +374,7 @@ def get_parameters(output_base, reform={}, test=False, baseline=False,
         output_base, 'TxFuncEst_{}.pkl'.format(guid),
     )
     if run_micro:
-        txfunc.get_tax_func_estimate(
+        txfunc.get_tax_func_estimate(S, 
             baseline=baseline, analytical_mtrs=analytical_mtrs,
             age_specific=age_specific, start_year=start_year,
             reform=reform, guid=guid, tx_func_est_path=tx_func_est_path
--- a/ogusa/parameters.py
+++ b/ogusa/parameters.py
@@ -33,7 +33,8 @@ import pickle
 import txfunc
 import elliptical_u_est as ellip
 import matplotlib.pyplot as plt
-from ogusa.utils import DEFAULT_START_YEAR
+from ogusa.utils import DEFAULT_START_YEAR, TC_LAST_YEAR
+from ogusa import txfunc
 
 
 '''
@@ -44,7 +45,6 @@ Set paths, define user modifiable parameters
 PARAMS_FILE = os.path.join(os.path.dirname(__file__), 'default_full_parameters.json')
 PARAMS_FILE_METADATA_NAME = 'parameters_metadata.json'
 PARAMS_FILE_METADATA_PATH = os.path.join(os.path.dirname(__file__), PARAMS_FILE_METADATA_NAME)
-TAX_ESTIMATE_PATH = os.environ.get("TAX_ESTIMATE_PATH", ".")
 USER_MODIFIABLE_PARAMS = ['g_y_annual', 'frisch', 'world_int_rate']
 DEFAULT_WORLD_INT_RATE = 0.04
 
@@ -130,7 +130,7 @@ def get_parameters_from_file():
     RETURNS: j
     --------------------------------------------------------------------
     '''
-    with open(PARAMS_FILE,'r') as f:
+    with open(PARAMS_FILE, 'r') as f:
         j = json.load(f)
         for key in j:
             if isinstance(j[key], list):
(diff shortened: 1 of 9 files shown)


ogusa/tests/test_startyears.py
```diff
@@ -0,0 +1,58 @@
...
+    input_dir = "./OUTPUT"
+    for year in range(2016, 2020):
+        print('year = ', year)
+        user_params = {'frisch':0.41, 'debt_ratio_ss':1.0,
+                       'start_year': year}
+        runner(output_base=output_base, baseline_dir=input_dir, test=True,
+               time_path=True, baseline=True, age_specific=False,
+               user_params=user_params, run_micro=True,
+               small_open=False, budget_balance=False, data=PUF_PATH)
+
```
hdoupe: @jdebacker Why did you choose years 2016 through 2020? Isn't the minimum year that can be specified 2013?

Also, it would be more efficient to just check the minimum and maximum years (something like 2013 and 2020) and just one or maybe two years in between. I don't think you would lose much if any test coverage if you were to do this. If, for example, 2013, 2017, and 2020 work, then I would expect 2014, 2015, 2016, 2018 and 2019 to work, too.

jdebacker: Good call @hdoupe - will do this.  Thanks.

ogusa/tests/test_startyears.py
```diff
@@ -0,0 +1,23 @@
...
+    output_base = "./OUTPUT"
+    input_dir = "./OUTPUT"
+    for year in [2013, 2017, 2026]:
+        print('year = ', year)
+        user_params = {'frisch': 0.41, 'debt_ratio_ss': 1.0,
+                       'start_year': year}
+        runner(output_base=output_base, baseline_dir=input_dir, test=True,
+               time_path=True, baseline=True, age_specific=False,
+               user_params=user_params, run_micro=True,
+               small_open=False, budget_balance=False, data=PUF_PATH)
```
hdoupe: @jdebacker you may want to use the [`pytest.mark.parametrize`](https://docs.pytest.org/en/latest/parametrize.html) decorator here. There are a couple reasons to use this:
1. if year 2013 fails, years 2017 and 2026 will still run.
2. you can use the [xdist py.test plugin](https://docs.pytest.org/en/3.0.0/xdist.html) to run parametrized tests in parallel. Tax-Calculator uses this. There's some explanation on how to use it in the TC [TESTING.md](https://github.com/open-source-economics/Tax-Calculator/blob/master/TESTING.md).

ogusa/parameters.py
```diff
@@ -130,18 +130,21 @@ def get_parameters_from_file():
...
-                   user_modifiable=False, metadata=False,
-                   tx_func_est_path=None, start_year=DEFAULT_START_YEAR,
-                   constant_rates=True, **small_open):
+def get_parameters(output_base, reform={}, test=False, baseline=False,
+                   guid='', user_modifiable=False, metadata=False,
+                   run_micro=False, constant_rates=True,
+                   analytical_mtrs=False, age_specific=False,
+                   start_year=DEFAULT_START_YEAR, data=None,
+                   **small_open):
+
```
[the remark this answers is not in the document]
jdebacker: @hdoupe It's a function of the `output_base` argument, so it's not pass in separately (it'll be created in `parameters.get_parameters()`.  

But let me look at that test and see what I need to change.  Thanks for pointing this out.

ogusa/tests/test_startyears.py
jdebacker: @hdoupe I'll look at doing this, (1) would be nice.  Re (2), I don't think that will work here as the tax function estimation done at each step already uses parallelization, so I think I'd run into the issue that subprocesses can't spawn their own subprocess.

hdoupe: Ok sounds good. 

(2) Ah, I keep forgetting about that.

jdebacker: @hdoupe See my edits to this test to incorporate `pytest.mark.parameterize`.  Appreciate the suggestion to do that.

ogusa/tests/test_parameters.py
```diff
@@ -58,8 +58,8 @@ def test_parameters_metadata_baseline():
     [('', None, "./TxFuncEst_{}.pkl"),
      ('', "test.pkl", "test.pkl"),
      (9, None, "./TxFuncEst_{}9.pkl")])
-def test_tx_func_est_path((monkeypatch, baseline, guid, tx_func_est_path,
-                          exp_tx_func_est_path)):
+def test_tx_func_est_path(monkeypatch, baseline, guid, tx_func_est_path,
```
jdebacker: @hdoupe I need some help fixing  this test.  I'm a bit confused by what it is trying to do as it's not testing directly a function in `parameters.py`.  Do I need some pickle files sitting in specific locations?

jdebacker:
@hdoupe I think this PR is ready, except that I can't figure out what I should be doing with the `test_parameters.test_tx_func_est_path()` test.  I feel like it might still be useful to have around, but not sure how this should be modified.  Let me know if you have suggestions.  Thanks!

ogusa/parameters.py
```diff
@@ -366,44 +370,39 @@ def get_parameters(test=False, baseline=False, guid='',
...
-        if tx_func_est_path is None:
-            policy_pckl = "TxFuncEst_policy{}.pkl".format(guid)
-            estimate_file = os.path.join(TAX_ESTIMATE_PATH,
-                                         policy_pckl)
-        else:
-            policy_pckl = "TxFuncEst_policy{}.pkl".format(guid)
-            estimate_file = tx_func_est_path
+        policy_pckl = "TxFuncEst_policy{}.pkl".format(guid)
+        estimate_file = tx_func_est_path
         print 'using policy tax parameters'
```
hdoupe: @jdebacker asked about `test_tx_func_est_path`
> @hdoupe I need some help fixing this test. I'm a bit confused by what it is trying to do as it's not testing directly a function in `parameters.py`. Do I need some pickle files sitting in specific locations?

I created that test to check the if statement logic that I added here. I wanted to check that the correct file paths were passed to `read_tax_func_estimate`. However, it looks like you are limiting the tax function parameter file paths to the form `"TxFuncEst_policy{}.pkl"`. So, I don't think the test is needed anymore.

jdebacker: Thanks for the suggestion @hdoupe.  I've removed the test.

PR #343, opened 2018-01-17
Tax array refactoring

jdebacker:
This PR updates `tax.py` to use more concise syntax in setting the tax function parameters.  It also updates the syntax used in `tax.py` to be more in compliance with PEP8 standards.

PR #349, opened 2018-02-07
[WIP] Tests household.py functions

jdebacker:
This PR sets up tests for the functions in the `household.py` module.


jdebacker:
@hdoupe I think some of the recent test failures in Travis CI are due to the time limit for tests being exceeded.  Obviously, I could think about pulling some tests out of what is run on Travis CI.  Do you recommend anything else (e.g., is there a way to increase the run time on Travis CI)?

hdoupe:
@jdebacker It looks like there is a [build time-out](https://docs.travis-ci.com/user/customizing-the-build/#Build-Timeouts) at 120 minutes which seems plenty long enough to run the core test suite. I just looked back at PR #343 and the tests took 20 minutes to run. Which tests that have been added since then are taking so long?

jdebacker:
@hdoupe Two long-running tests were added: 

* `test_startyears.py`
* `test_user_inputs.py`

However, I used the decorator `@pytest.mark.full_run` for each.  I was hoping that meant that they wouldn't only be run if one specified the full run in `pytest`.  But they must be running in the Travis CI tests.  How do I avoid this?

PR #350, opened 2018-02-23
Test SS.py

jdebacker:
This PR adds tests of the functions in `SS.py`.


jdebacker:
@hdoupe Do you think it is ok to use pickle files to have cached the inputs and outputs for testing as I am doing here?  I've had it recommended that it might not be a good idea to have binary files in the repo, but there are so many objects of different types, that if I try to save inputs/outputs to JSON, the many objects (e.g., all the Numpy arrays) do not have their type preserved.

hdoupe:
@jdebacker I haven't had much experience with pickling files in general. I usually try to save the data in some kind of plain-text format like JSON or CSV. My main concern with saving pickled data is that there may be issues with reading it in both Python 2 and Python 3. Or, there could be an issue if the NumPy or Pandas version changes, etc.

However, I also understand that it is convenient to load and unload NumPy objects. Perhaps, there is some standard way to serialize the numpy data before you write it to the file.

jdebacker:
@hdoupe Thanks for your comments.  The Python 2 vs 3 issue is important - and why I used JSON for a similar caching of data for B-Tax.  But since OG-USA is run only on 2.7 (and I hope jump straight to Python 3.6 without having to be compatible with both), I used pickle's here.

You can serialize a Numpy array in JSON, but the issue is that I wanted to cache a large tuple of parameters in one file.  If I dump it all to JSON, then I can't just load that JSON and have my original datatypes back.  What were Numpy arrays will not be lists and I'll have to cast each one as a Numpy array again after loading.

Unless you have a real problem with it, I'm inclined to go with the pickle format for now then try to use JSON in the future.  My priority is to get *some* testing of these functions up ASAP since I have a backlog of things I want to do to add to OG-USA, but want to ensure they don't change output.

hdoupe:
@jdebacker That makes sense to me. Using the pickle sounds like a good idea. At some point in the future, we can take another look at this problem.

ogusa/tests/test_SS.py
```diff
@@ -0,0 +1,163 @@
...
+                           'run_SS_open_unbal_outputs.pkl'),
+                          ('run_SS_closed_balanced_inputs.pkl',
+                           'run_SS_closed_balanced_outputs.pkl')],
+                         ids=['Open, Unbalanced', 'Closed Balanced'])
+def test_run_SS(input_path, expected_path):
+    # Test SS.run_SS function.  Provide inputs to function and
+    # ensure that output returned matches what it has been before.
+    input_tuple = pickle.load(open(os.path.join(
+        CUR_PATH, 'test_io_data', input_path), 'rb'))
+    (income_tax_params, ss_params, iterative_params, chi_params,
```
hdoupe: You may want to do:
```
with open(...) as f:
    input_tuple = pickle.load(f)
```

You get some nice behavior regarding error handling and closing the file connection once the data has been read.

jdebacker:
This PR now fixed a bug introduced to `get_micro_data.py` in PR #351.  Start years from 2013-2016 should now run without issue.

I've had to narrow the upper bounds of the Frisch and growth rate parameters that the model can handle.  I'm not sure why the tests weren't passing with the previous range, but we should look out for this.  My hunch is that the need to change the upper bound of the Frisch elasticity parameter change might have come from the addition of the stitching function in the `household.marg_ut_lab()` function.  But I think these issues of the model not running with values outside this relatively narrow range are due to the poor starting values we have (I've run this model with Frisch values of 1.5+ with different starting values).  Future PRs will improve these starting values and robustness.  Just need more testing  before doing that.

cc @rickecon

Issue #352, opened 2018-03-02
PolicyBrain API


hdoupe:
This issue outlines an interface that would ease the integration of OG-USA onto PolicyBrain:

1. JSON document containing model specifications (see [`taxcalc/current_law_policy.json`](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/current_law_policy.json) for examples)
    a. long name, parameter description to be displayed in the hover over box on PB, needs default value, type hints--`integer_value` or `boolean_value` (assumed to be float otherwise), range : `{'min': min value, 'max': max value}`, action on validating range--either `stop` or `warn`, warning message on violating
2. `default_data` -- returns a python dictionary containing the contents of the JSON document containing the model specifications (see [`taxcalc/parameters.default_data`](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/parameters.py#L23-L52))
3. `read_json_param_objects`: final specification formatting (see [`taxcalc/calculator.read_json_param_objects`](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/calculate.py#L1023-L1096))
    a. PoliyBrain converts this dictionary `{'param_name': 'raw input data', ...}` to `{'param_name': {'YEAR_i': 'parsed val_i', 'YEAR_j': 'parsed val_j', ...}, ...}`. This dictionary is passed to `read_json_param_objects` for final formatting
4. `reform_warnings_errors`: provides error and warning messages given the parsed user input
  a. takes input from `read_json_param_objects` and returns dictionary `{'errors': 'new line separated error messages', 'warnings': 'new line separated warning messages'}` (see [`taxcalc/tbi/tbi.reform_warnings_errors`](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/tbi/tbi.py#L42-L77))

@jdebacker , if you can open up a PR for creating the JSON specifications object, we can work out how the different specifications can be mapped to the OG-USA parameters. Once this is set up, we can take a look at extending the `taxcalc/parameters.py` module to work with this document. My hunch is that once we get that connection setup the final formatting and warning/error handling will fall into place as we can borrow parts of the taxcalc warning and error handling code.

Also, the API endpoint names are flexible. These are just taken from Tax-Calculator for simplicity.

martinholmer:
@hdoupe said in issue #352:
> This issue outlines an interface that would ease the integration of OG-USA onto PolicyBrain

[...]

PR #354, opened 2018-03-09
PolicyBrain API

jdebacker:
This PR attempts to bring OG-USA into line with the PolicyBrain API using Issue #352 as a guide.

### DIFF
--- a/ogusa/default_parameters.json
+++ b/ogusa/default_parameters.json
@@ -1,0 +1,32 @@
+{
+    "frisch": {
+        "long_name": "Frisch elasticity of labor supply",
+        "description": "Frisch elasticity of labor supply",
+        "notes": "See Altonji (JPE, 1996) and Peterman (Econ Inquiry, 2016) for estimates of the Frisch elasticity",
+        "col_var": "",
+        "col_label": "",
+        "boolean_value": false,
+        "integer_value": false,
+        "value": [0.4],
+        "range": {"min": 0.02, "max": 0.06},
+        "out_of_range_minmsg": "",
+        "out_of_range_maxmsg": "",
+        "out_of_range_action": "stop",
+        "compatible_data": {"puf": true, "cps": true}
+    },
+    "g_y_annual": {
+        "long_name": "Exogenous growth rate of labor augmenting technological change",
+        "description": "Growth rate of labor augmenting technological change",
+        "notes": "",
+        "col_var": "",
+        "col_label": "",
+        "boolean_value": false,
+        "integer_value": false,
+        "value": [0.03],
+        "range": {"min": -0.01, "max": 0.04},
+        "out_of_range_minmsg": "",
+        "out_of_range_maxmsg": "",
+        "out_of_range_action": "stop",
+        "compatible_data": {"puf": true, "cps": true}
+    },
+}


hdoupe:
@jdebacker This looks good so far. I quickly sketched out how I think an OG-USA parameters class would look on this [branch](https://github.com/hdoupe/OG-USA/blob/pb_api/ogusa/pb_api.py). It will need to be filled in and integrated with OG-USA, but for the purposes of reasoning about the API, I think it should do.

What do you think? Do you think that type of approach could work?

cc @martinholmer

jdebacker:
@hdoupe Thanks for your work in the other branch.  I have a number of questions.  The first two are:

1) Do I need to put all parameter values currently specified in [parameters.py](https://github.com/open-source-economics/OG-USA/blob/master/ogusa/parameters.py) into the default_parameters.json file?
2) I'm not sure where in the process of running OG-USA the pb_api.py script you created will come into play?  We might need to set up a time to chat to help me understand this.

hdoupe:
I think moving the hard-coded parameters into the `default_parameters.json` document is a good idea. In the future, we could think about splitting it up into different categories like Tax-Calculator's `consumption.json`, `growdiff.json`, etc.

I think that instead of doing all of the parameter validation and processing in places like [this in `execute.py`](https://github.com/open-source-economics/OG-USA/blob/master/ogusa/scripts/execute.py#L30-L153) or [like this in `parameters.py`](https://github.com/open-source-economics/OG-USA/blob/master/ogusa/parameters.py#L359-L371) we could do it in `pb_api.py` or in a new, refactored `parameters.py`. 

A call sounds good--just sent you a message to set that up.

ogusa/model_solution_parameters.json
```diff
@@ -0,0 +1,152 @@
...
+        "col_label": "",
+        "boolean_value": false,
+        "integer_value": false,
+        "value": [None],
+        "range": {"min": , "max": },
+        "out_of_range_minmsg": "",
+        "out_of_range_maxmsg": "",
+        "out_of_range_action": "stop",
+        "compatible_data": {"puf": true, "cps": true}
+    },
```
hdoupe: @jdebacker how is `MINIMIZER_OPTIONS` used? I did a search in the code base and can see where it's set but not where it is used.

jdebacker: @hdoupe You are right - it is not used.  I'll drop this from the JSON.

hdoupe: Ok, sounds good.

jdebacker:
Deleted unused `MINIMIZER_OPTIONS` from JSON.  

Added @hdoupe's work on `pb_api.py` -- with some additional called to cheap parameter computations.

hdoupe:
@jdebacker this looks good so far.

I put some thought into some of the questions you posed:
> How can I:
> i) check things like that lambdas.sum=1?  Length J
> ii) what to do with world_int_rate dictionary?
> iii) Checking tG1 vs tG2 vs T
> iv) Check that tau_bq is of length J
> v) How to choose set of test parameters (small state space) easily
> vi) Checking consistency of baseline_spending with other options
> vii) how handle dask parameters?
> viii) How handle baseline boolean?  paths? 

i, iv, vi) We may have to do the initial range validation and then do further validation on these more complex requirements.

v) Is this something that we could specify with the parameters in the JSON files? If so, could you just create a JSON file like the Tax-Calculator presets that specifies the test parameters?

vii) Would it make sense to include those in the `model_solution_parameters.json` file?

viii) I feel like the baseline boolean parameter could be included in the `default_parameters.json` file? However, it seems like it might make more sense to keep that as a normal variable that is passed along and not specified in a JSON file. But, I'm not really sure. I don't think the paths necessarily have to be specified in a JSON file either.

iii) I was able to get a working solution for this which allows you to set the range for a parameter like:
```
"range": {"min": f(other_parameter_name, constant), 'max': g(other_parameter_name, constant)}
```
where f or g is either:
value of other_parameter_name + constant
value of other_parameter_name - constant
value of other_parameter_name * constant
value of other_parameter_name / constant

All of this is on my [ogusa-paramsbase](https://github.com/hdoupe/OG-USA/tree/ogusa-paramsbase) branch. The function that extends the range capabilities is [here](https://github.com/hdoupe/OG-USA/blob/ogusa-paramsbase/ogusa/parametersbase.py#L370-L402) and it's called from [here](https://github.com/hdoupe/OG-USA/blob/ogusa-paramsbase/ogusa/pb_api.py#L213). I need to spend some time cleaning up the code, documenting, and adding a test or two. 

Then, I could open up a PR on this branch that you are working on. Opening a PR on a PR is kind of a clumsy way to integrate changes, but I don't know of a better way to do it. If you want to try another process, then we can talk about that.

hdoupe:
See jdebacker/dynamic#9 for further implementation of the `pb_api.py` module in this PR.

asmockler:
@jdebacker Anything I can do to help move this along? Or are there things listed in #352 that I could do independent of the work happening here?

jdebacker:
@asmockler As you can see, my work on this branch has been intermittent and I just set out the basics of the parameter defaults and so on as outline by @hdoupe.  

I feel like I need to think more about how OG-USA needs to be restructured to work better with the PolicyBrain API and forthcoming changes to PB.  At the moment, I'm having hard time thinking of places to make small independent contributions because I don't have an exact idea what the final structures should look like.  Let's put any discussion over in Issue #352 or new issues.

PR #360, opened 2018-03-16
Dask for multiprocessing in tax function estimation

jdebacker:
This PR will replace the multiprocessing library with dask for the parallel computing of the tax function estimation routine.

This will be a benefit when doing distributed computing.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -14,3 +14,4 @@ dependencies:
 - pandas>=0.20.1
 - pytest
 - python=2.7
+- dask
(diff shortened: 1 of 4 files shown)


jdebacker:
I'm getting serval warning messages and some odd behavior with the object returned from `compute`.  These were not present when testing `txfunc.tax_func_estimate()`.   Need to look into why these would arise when calling this function from others.

hdoupe:
@jdebacker what do the warnings look like?

Also, I think you need to add dask to the `environment.yml` file?

jdebacker:
@hdoupe The warning is:

```
distributed.core - WARNING - Event loop was unresponsive in Worker for 1.35s.  This is often caused by long-running GIL-holding functions or moving large chunks of data. This can cause timeouts and instability.
```

This appears if using `num_workers`>1.  Despite the warnings, the function runs in parallel and the results make sense (i.e., match what I get in serial or in parallel with multiprocessing).

PR #362, opened 2018-03-20
Parallelize SS solution

jdebacker:
This PR implements a parallelized solution to the model's steady state using dask.


jdebacker:
Note that some of the test data needed to be updated.  The parallelized solution produces the same results for the SS, but because several tests (`test_inner_loop()`, `test_SS_fsolve()`, `test_SS_fsolve_reform`, and `test_SS_fsolve_reform_baselinespending()`) only solve for one iteration of the solution algorithm (i.e., they do not solve until convergence), starting values matter.  In the serial solution method, starting values for the problem for the type `j` agents where taken from the solution for type `j-1` agents.  This can't be done when solving for the `j` types in parallel.  Thus the results after one iteration are different in the parallelized method.

hdoupe:
@jdebacker said: 
>  In the serial solution method, starting values for the problem for the type j agents where taken from the solution for type j-1 agents. This can't be done when solving for the j types in parallel. Thus the results after one iteration are different in the parallelized method.

What if you did something like this:
1. solve for agent `J=1` in serial
2. solve for agents `J={2: 2 + num_workers}` using the results from `J=1` as the starting values
3. solve for agents `J={2 + num_workers + 1: 2 + 2 * num_workers + 1}` using the results from` j = 2 + num_workers`
4. repeat until done

Then, you could still run the solver in parallel, and maybe have better starting values.

Does that make sense?

jdebacker:
@hdoupe Your algorithm makes sense.  Let me think about whether/how to implement.  What I've found in some other work is that using the `j-1` solution as starting values is generally not as good as using the `j` values from the prior iteration through the steady state.  The only issue is getting the best guess at starting values for each `j` in the first iteration through the  SS/TPI solution algorithms.  I intend to think more about the best way to find good starting values in another PR.

hdoupe:
Ok, I see. That makes sense. Thanks for the explanation.

PR #367, opened 2018-04-02
Python 3.6 compatibility

jdebacker:
This PR updates OG-USA to run on both Python 2.7 and Python 3.6.

cc @rickecon


jdebacker:
All tests except for pickle comparisons in `test_basic.py` are passing in Python 3.6.

However, running the model generates errors.  It looks like the solver for the household problem in the SS goes out of bounds and stays there, generating `nan`'s.  This doesn't happen in the tests of `SS.py`, so it is likely related to some input to the model changing.  This needs to be diagnosed.

jdebacker:
Update: error running model has to do with trying to run model with analytical mtrs with tax function estimates with constant rates - this doesn't work given how we implement constant rates (because you get a division by zero issue in the mtr functions in `tax.py`.

hdoupe:
@jdebacker what do you think about moving the try/except logic into a function?

Then, you could do something like:

```
def safe_read_pickle(file_path):
    with open(file_path, 'rb') as f:
        try:
            obj = pickle.load(f, encoding='latin1')
        except TypeError:
            obj = pickle.load(f) 
    return obj
```

That way you don't have so many of the try/except clauses in the code base. This would also make changing the logic around reading pickle files easier in the future since you would only have to change it in the function and not a bunch of times throughout the code-base.

jdebacker:
Ok - I'm going to incorporate @hdoupe's suggestion to simplify the code a bit.  

@rickecon, I think PB is still running 2.7.  So I think we go ahead an maintain 2.7 compatibility for OG-USA for now (since it already has it).  We can then simplify the code by dropping lines necessary only for backward compatibility in a future PR (around the time PB switches over).

environment.yml
```diff
@@ -13,5 +11,6 @@ dependencies:
 - numpy>=1.12.1
 - pandas>=0.20.1
 - pytest
-- python=2.7
```
asmockler: I think it would be nice to declare `python=3.6` as a dependency to encourage people to use it moving forward. Might make it easier to fully drop `2.7` when the time comes

Issue #92, opened 2015-07-17
Consolidate Graphing?


evan-magnusson:
Currently, we have 2 main graphing files, SS_graphs.py and TPI_graphs.py.  However, we also do some graphing in demographics.py, income.py, wealth_data.py, and labor_data.py.  These other graphing functions are controlled by a flag, but perhaps it might be best to move all of these graphing functions into SS_graphs.py.  That way, all graphing occurs in one place.
     This should be pretty easy for income.py, wealth_data.py, and labor_data.py, since what they graph is saved anyway.  However, for demographics.py, some of the objects it graphs are not currently saved.  This is easily changed, but it will need to change in order for another separate function to graph the fertility rates and immigration rates, for example.
    This isn't terribly important, but might be something to think about in a later version.

jdebacker:
@rickecon How to you feel about this suggestion?

I like it because it:
1) Reduces the lines of code in any script.  Pulling out the graphing commands will make it easier to read the executable code.
2) By putting all the graphs in one place, we might be more likely to reduce the lines of code by doing things like creating a [style sheet](https://matplotlib.org/users/style_sheets.html) or making a plotting function flexible so that it can handle multiple inputs.
3) We've talked about something related to this recently - having a script to produced more detailed output (tables and figures) from OG-USA runs (such as the script you created for the QNs).

This will take some time to map out what objects will need to be saved, etc., but I think it could be helpful.

jdebacker:
Not quite 4 years later, I plan to get started on this.   The outline is the following:

Have 4 different modules:
1. OG-USA parameter plots (takes parameter class as argument)
  * Plot of lifecycle earnings profiles
  * Plot of initial and adjusted immigration rates
  * Plot of initial and SS population distribution (and maybe some along the way)
  * Plot of mortality rates by age
  * Plot of fertility rates by age
  * Plot pop growth rate by year with ss value
  * Plot of elliptical utilty vs constant frisch
    * Plot can do marginal or levels of utility
  * Plot ETR/MTRs by income
  * 3D plot of surface of tax functions with scatter plot of ETR data
  * 3D scatter plot of tax function data
2. OG-USA parameter tables (takes parameter class as argument)
  * Table with all model parameters
  * Mean ETR/MTRs over budget window
3. OG-USA output plots (takes paths to output as arguments, maybe also number of periods)
  * Plot percentage change in macro vars (several together - Y, L, K or I, G, C)
[...]

PR #378, opened 2018-04-26
Remove unused Saved_moments directory

asmockler:
Resolves#370

This PR removes references to the `Saved_moments` directory that was being created but never used.

I'm not sure if there are cases in which this will break the graph scripts – it is reading in data and dumping it into the global namespace and I don't think there is a good way to track down all the side-effects that would have had 🤔 

@jdebacker @rickecon

### DIFF
--- a/ogusa/SS_graphs.py
+++ b/ogusa/SS_graphs.py
@@ -10,10 +10,6 @@ This py-file calls the following other file(s):
             household.py
             SSinit/ss_init_vars.pkl
             SS/ss_vars.pkl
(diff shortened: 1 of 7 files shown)


jdebacker:
@asmockler Thank you for this PR.  That `SS_graphs.py` script is a mess.  I think we need to do some improvements on how to handle the plotting of output in general (see issue #92).  But I think this will be best addressed after we make some updates work with the PB API (see PR #354).

Merging.

PR #379, opened 2018-04-27
Remove use of pickle in model runs

jdebacker:
This PR replaces the pickle files generated during a run of OG-USA with json files.  This is done to ensure 1) improved compatibility across different version of python and  2) the ability to read these objects into other software if necessary.

Note that this PR does not help with removing pickle files that have been checked into the repo for unit tests.


jdebacker:
@hdoupe, I had not heard of pyarrow but I will look into it.  Sounds like a great option for the unit testing data.   Thanks for the suggestion.

hdoupe:
@jdebacker Here's another cool development for saving arrays and preserving metadata/data types: https://pandas.pydata.org/pandas-docs/version/0.23.0/whatsnew.html#json-read-write-round-trippable-with-orient-table

Looks like you can dump a pandas DataFrame to a JSON file and read it back without losing any metadata information. This is a pandas pre-release. So, the feature won't be available for some time. I think it's something to keep an eye on.

jdebacker:
Closing this PR due to inactivity.  Still a goal to remove pickle files from the repo, especially those associated with testing, which are adding a lot of bloat to the `ogusa` package.

PR #401, opened 2018-10-19
Tests for demographics.py

jdebacker:
This PR introduces tests of the `demographics.py` module.  These test:
1. That the population distribution at the end of the time path is the same as in the SS.
2. That the population distribution evolves "smoothly" (i.e., no large jumps from one year to the next).
3. That the population growth rate evolves "smoothly" (i.e., no large jumps from one year to the next).

### DIFF
--- a/ogusa/SS.py
+++ b/ogusa/SS.py
@@ -366,7 +366,7 @@ def inner_loop(outer_loop_vars, params, baseline,
                                                args=euler_params,
                                                xtol=MINIMIZER_TOL,
                                                full_output=True))
-    results = compute(*lazy_values, get=dask.multiprocessing.get,
+    results = compute(*lazy_values, scheduler=dask.multiprocessing.get,
                       num_workers=num_workers, scheduler=client)
 
     # for j, result in results.items():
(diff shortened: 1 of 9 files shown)


jdebacker:
@hdoupe Any idea how to fix this Dask issue causing the failure?  Guessing it has to do with Travis CI using a different (probably newer) version of Dask than I am running locally, because I'm not getting these failures.

hdoupe:
Hmm, that seems about right to me. Where is the `Client` created?

jdebacker:
For these tests, it just takes the default `client=None`

hdoupe:
I dropped the `get` key word from: 
```
# SS.py line 370
    results = compute(*lazy_values, get=dask.multiprocessing.get,
                      num_workers=num_workers, scheduler=client)
```

so that now it's:

```
# SS.py line 370
    results = compute(*lazy_values,
                      num_workers=num_workers, scheduler=client)
```

and that seems to work. It looks like the current behavior is to use processes, but I'm not sure how to specify that from the new API.

jdebacker:
Thanks @hdoupe.  I updated my Dask version and also found that `scheduler=dask.multiprocessing.get` works.  Will update this PR soon - running some tests on another branch now.

hdoupe:
No problem!

Issue #414, opened 2018-10-23
Distributed memory issue: OSError: [Errno 12] Cannot allocate memory


rickecon:
I ran `run_ogusa_example.py` yesterday on my Dell Precision 7820 desktop with the following specs:

* Ubuntu Linux 16.04
* Intel Xeon Gold 6138 2.0GHz, 3.7GHz Turbo, 20C, 10.4GT/s 3UPI, 27M Cache, HT (125W) DDR4-2666 1st
* Intel Xeon Gold 6138 2.0GHz, 3.7GHz Turbo, 20C, 10.4GT/s 3UPI, 27M Cache, HT (125W) DDR4-2666 2nd
* NVIDIA Quadro P1000, 4GB, 4 mDP (7X20T)
* 32GB (4x8GB) 2666MHz DDR4 RDIMM ECC
* M.2 512GB PCIe NVMe Class 40 Solid State Drive
* 2.5" 1TB 7200rpm SATA Hard Drive

So this machine has 40 cores (80 threads), 32GB Memory, an OK GPU, and a big hard drive.

When I ran OG-USA, I got the following error, `OSError: [Errno 12] Cannot allocate memory`. This is during the estimation of the tax functions in the baseline (inside of the `get_micro_data.py`, `get_data()` function). 
```
(ospcdyn) rickecon@evans50core:~/Documents/Economics/OG-USA/run_examples$ python run_ogusa_example.py
You loaded data for 2014.
Tax-Calculator startup automatically extrapolated your data to 2014.
Number of workers =  80
making dir:  ./OUTPUT_BASELINE/SS
making dir:  ./OUTPUT_BASELINE/TPI
In runner, baseline is  True
BW =  10 start year =  2018
BW =  10 begin year =  2018 end year =  2027
Running current law policy baseline
You loaded data for 2014.
Tax-Calculator startup automatically extrapolated your data to 2014.
/home/rickecon/anaconda3/envs/ospcdyn/lib/python3.6/site-packages/dask/base.py:829: UserWarning: The get= keyword has been deprecated. Please use the scheduler= keyword instead with the name of the desired scheduler like 'threads' or 'processes'
    x.compute(scheduler='threads') 
or with a function that takes the graph and keys
    x.compute(scheduler=my_scheduler_function)
  "The get= keyword has been deprecated. "
year:  2020
year:  2023
year:  2026
year:  2021
year:  2022
year:  2025
year:  2027
year:  2024
[...]

PR #415, opened 2018-10-23
Updated execute.py with tG1 and tG2 user_params, and cleanup

rickecon:
1. The main thing this PR does is updates `execute.py` in the `scripts` directory with allowing the `user_params` to include `tG1` (year in which budget closure rule starts) and `tG2` (year in which budget closure rule hits target. I ran this with `tG1=32` (year 2050) and `tG2=42` (year 2060). Interestingly, those pushed out closure years resulted in negative steady-state government spending for a target steady-state D/Y of 1.0. This is interesting because it is a measure of policy infeasibility.
2. The second minor thing this PR does is to pull import statements out of functions in `execute.py` and put them at the top of the module. It can be a memory sink to import modules every time that a function is called. However, I tested this change and it was not the memory sink that was causing Issue #414.
3. Lastly, I cleaned up some PEP8 violations :)

### DIFF
--- a/ogusa/scripts/execute.py
+++ b/ogusa/scripts/execute.py
@@ -13,12 +13,11 @@ import numpy as np
 import time
 
 import ogusa
-from ogusa import calibrate
+from ogusa import calibrate, parameters, demographics, income, utils, SS, TPI
 from ogusa.parameters import DEFAULT_WORLD_INT_RATE
 ogusa.parameters.DATASET = 'REAL'
 from ogusa.utils import DEFAULT_START_YEAR, TC_LAST_YEAR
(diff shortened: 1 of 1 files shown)


jdebacker:
LGTM.  @rickecon is this ready to go?

As an aside, other than running the regression tests, the `open-source-economics:dev` branch is ready to be merged into `open-source-economics:master`.  This should make it a lot easier to adjust parameters that were not previously in `user_params` without having to add each individually as this PR does. I'm about 95% sure the regression tests should be ok given other tests passing, but want to be sure to run those since it's such a major refactoring.  I just haven't found compute time elsewhere and don't want to tie up my laptop for a few days running them.

PR #420, opened 2018-11-09
Move dev branch with parameters class into master

jdebacker:
This PR is to move the `open-source-economics:dev` branch into `open-source-economics:master`.  The code has passed all unit and regression tests on the dev branch.

A few changes will need to be made to `test_demographics.py` to pass this new unit test that is in the `master`, but not the `dev` branch.

PR #421, opened 2018-11-09
Update master branch to use new parameters class

jdebacker:
This PR replaces PR #420.  It fixed the conflicts that prevented that PR.

This represents a major refactoring of the OG-USA codebase to use a parameters class that has the following benefits:
1. Parameter validation
2. A default parameter JSON that provides metadata and contains all input parameters in one place.
3. Allows all model parameters to be easily updated.
4. Moves towards an API that will work well with PSL.
5. Removes the passing of parameters through tuples to many functions, which was error prone, added a significant number of lines to the code, and made it more difficult to update modules (since one had to trace the passing of these tuples through several modules).


jdebacker:
@hdoupe I was thinking about merging before renaming, but happy to rename in this PR.  

The `parameters.py` module has been removed - so I think it's ok to to rename `pb_api.py` to that.  Should it be `parameters.Specifications()` as the class or  `parameters.Parameters()`?  I'm leaning towards the former and keeping the `Specifications` since if we rename the module it's in will make it obvious that its a specification of the parameters.

hdoupe:
@jdebacker said:
> @hdoupe I was thinking about merging before renaming, but happy to rename in this PR.

Ah, that works, too.

> The parameters.py module has been removed - so I think it's ok to to rename pb_api.py to that. Should it be parameters.Specifications() as the class or parameters.Parameters()? I'm leaning towards the former and keeping the Specifications since if we rename the module it's in will make it obvious that its a specification of the parameters.

Ok, I see. I think `parameters.py` is a good name for the module. I am fine with `Specifications` or `Parameters`. I think I'm leaning towards `Specifications`, too.

rickecon:
@jdebacker Should I close PR #420?

jdebacker:
@rickecon I think we can close it.  I was waiting to have this merged first though.  I think it's ready to go.

Issue #400, opened 2018-10-19
Retirement age over the time path


jdebacker:
I believe there is a bug in the time path solution related to the retirement age for cohorts with incomplete lifetimes.  In particular, the parameters `retire` is an integer representing the year of life beyond which agents are eligible for retirement benefits.  Over the these incomplete lifetimes, we have `retire` fixed at the same integer, even though the length of the lifetime varies.

This comes into play in the `tax.total_taxes()` function.  Here we have:

```
retireTPI = (retire - S)
...
T_P[retireTPI:] -= theta[j] * w[retireTPI:]
```

Because `retire` < `S` this never raises an indexing issue, even though sometimes the absolute value of `retire-S` exceeds the length-s vector that is `w` (e.g. you might have `w` and a vector of length 20 and `retire-S` = -40 so that you have `w[-40:]` which is still an array of length 20.

I think what we need is `retireTPI = (s-retire)`, where `s` is the number of periods of life remaining in the incomplete lifetime.

@rickecon can you confirm that this is an issue?

jdebacker:
The `retireTPI = (retire - S)`, actually gives the behavior that is intended (even if it seems weird to be indexing beyond the length of the vector.

PR #422, opened 2018-11-11
Changes to how retirement benefits are computed

jdebacker:
This PR aims to do 3 things:
1) Replace hardcoded values for parameters used to compute replacement rates with flexible parameters.
2) Allow users to specify a retirement age in calendar years (as opposed to model periods) by making this a parameter.
3) Fix the bug described in Issue #400, where the period of retirement is not handled correctly for some cohorts along the time path.


jdebacker:
Removing point #3 since I was mistaken in thinking this was a bug as noted in Issue #400.

PR #424, opened 2018-11-16
Updates to firm.py for general CES production

rickecon:
This PR updates the `firm.py` module to accommodate general CES production function. It also cleans up some of the docstrings. The theory behind these changes is somewhat involved and is characterized in the OG-USA documentation in Appendix A-1.

This PR also changes the minimum and maximum of the `epsilon` parameter in the `default_parameters.json` file from `[0, 1]` to `[0.2, 10.0]`. @jdebacker


jdebacker:
@rickecon Thanks for opening this PR - getting the general CES to work is a great contribution.

I see some failures of `tests/test_firm.py` that are due to test cases with `epsilon = 0.0`.  Looks like there might be some other places to check too (e.g., in how `Y` is computed).

It might be useful to add some tests for the K/L ratio function also.

rickecon:
@jdebacker . I'll work on these. I did remove `epsilon=0` as a possibility. I reset the range in `default_parameters.json` to `[0.2, 10]`.

jdebacker:
> I did remove epsilon=0 as a possibility. I reset the range in default_parameters.json to [0.2, 10]

Yes - I think that is a good fix.  But `tests/test_firm.py` has some cases with `epsilon = 0.0` that need to be changed (though if the `Specifications` class is working as expected the errors that are thrown should be different than what I'm seeing in the Travis CI results).

jdebacker:
Actually, I see why the `Specifications` isn't throwing the error I expected with these tests:

```
    p = Specifications()
    new_param_values = {
        'Z': 0.5,
        'gamma': 0.5,
        'epsilon': 0.0
    }
    # update parameters instance with new values for test
    p.update_specifications(new_param_values, raise_errors=False)
```
We should probably remove the `raise_errors=False` so that it throws and error with out of bounds parameters values such as `epsilon=0.0`.

rickecon:
@jdebacker. I am almost done with an update to `test_firm.py` that should allow this PR to pass CI tests.

jdebacker:
This looks good, thanks @rickecon

PR #437, opened 2019-01-07
Allow replacement rates to be adjusted along the time path

jdebacker:
This PR adds two features to the model:
1.  A scale parameter to adjust the social security retirement income replacement rate along the transition path.
2. Parameters that represent the risk premia on private debt by providing a haircut to the interest rate the  government pays.

Re #2, we'll need to think about how to represent the household's portfolio of private capital and government debt when their yields differ.


rickecon:
@jdebacker. We could put a little Cobb-Douglas preference aggregator into the household budget constraint where the savings amount `b_{s+1,t+1}` used to be. Let the choice of how much debt be `d_{s+1,t+1}`, let the interest rate on private capital be `r_t`, and let the haircut interest rate on government debt (savings) be` r_{g,t}`. Then the household budget constraint is (I dropped the `j` subscripts in everything but the market clearing conditions):

hhbc

This results in an adjusted private capital savings Euler equation:

eulb

and a demand for government debt that is a fraction of private capital savings.

dst

Then the capital market clearing equation becomes two capital market clearing equations,

mc

and the government debt law of motion and closure rules change to have `r_{g,t}` in them. This leaves us still with just one really endogenous interest rate `r_t`, and then a government interest rate `r_{g,t} that is a fairly trivial function of the private interest rate and a government debt amount that is a function of `alpha_d`, the two interest rates `r_t` and `r_{g,t}` and the savings amount `b_{s+1,t+1}`.

It feels like we should probably open a new PR for this one.

jdebacker:
The full test suite passes locally.  This is ready for review @rickecon 

Not that while the parameters to scale the gov't interest rate were added to the `default_parameters.json`, not adjustments were made to the model in this regard.  This PR simply added a parameter to adjust the replacement rate (`theta`) over the time path (e.g., as we did for the recent OG Symposium).

PR #438, opened 2019-01-14
Bequest transmission process

jdebacker:
This PR adds a new parameter, `zeta`, that represents a "bequest transmission process".  Each element of the `S`x`J` matrix that is `zeta` represents the fraction of total bequests that goes to agents of a given age (`s`) and ability type (`j`).

The PR will add the ability to distribute bequests according to `zeta`, while retaining the ability to distribute bequests uniformly within ability type (i.e., the current method).


jdebacker:
All tests passing locally, but `test_SS.test_inner_loop()` failing on Travis CI.  This is the only test that required a new pickle file that holds the output results.  I am not sure if there is some issue related to that.

jdebacker:
Adding the guess of BQ to the outer loop in the SS solution algorithm, does not affect the robustness of the model to different values of the Frisch elasticity.  Increasing the initial guess of `r_ss` from 0.05 to 0.09 allows one to solve the baseline model with a Frisch up to about 0.8.

jdebacker:
Ignoring this failing test for now. Will make sure passed on next PR, but again, it passes locally.  Can't tell if any issues reading the new pickle file for Travis CI.

rickecon:
Yes!

PR #439, opened 2019-01-24
Consumption taxes

jdebacker:
This PR adds a consumption tax parameter, with flexibility to vary across age, ability type, and time.


jdebacker:
@hdoupe I've got one Travis CI test failing in this PR.  In this particular test, the outputs the function is tested against come from a tuple stored in a pickle file.  I had updated this pickle file in a prior PR, but couldn't get Travis CI tests to pass on that PR (#438) even though they passed locally.  I though it might be that the tests were not testing against the pickle file which was modified in the PR but rather the one checked into the master branch.   But now the pickle file is checked into the master branch and the same test is still failing in Travis CI (but passing locally) for this PR.  

I'm at a loss about how I remedy this - or even to see where the error could be coming from.  Any ideas about how I compare these pickle files?  Why a diff for `pytest` on my machine vs on Travis CI?

hdoupe:
@jdebacker I just took a quick look at the error log. It looks like the values are all very close to each other:

```
        for i, v in enumerate(expected_tuple):
>           assert(np.allclose(test_tuple[i], v))
E           assert False
E            +  where False = <function allclose at 0x7fc5ddf32488>(array([[-5.32907052e-15,  1.28785871e-13, -1.43884904e-13, ...,\n         7.94151487e-02, -1.99840144e-15, -2.38697950e...  [ 2.46469511e-14,  0.00000000e+00, -4.06341627e-14, ...,\n         2.33635266e-03,  8.88178420e-16,  4.21884749e-15]]), array([[-3.55271368e-15,  1.39444012e-13, -1.44773082e-13, ...,\n         7.94153633e-02, -2.88657986e-15, -3.60822483e...  [ 2.48689958e-14, -2.22044605e-16, -3.97459843e-14, ...,\n         2.33633743e-03, -5.55111512e-15, -1.55431223e-15]]))
E            +    where <function allclose at 0x7fc5ddf32488> = np.allclose
tests/test_SS.py:291: AssertionError
```

Are there one or more diff values that are substantially different from zero?

hdoupe:
@jdebacker The `test_SS.py::test_inner_loop` test passed locally for me too. Do you think that this is another tolerance issue?

jdebacker:
@hdoupe Thanks for the helpful references -- and for the suggestion to adjust the tolerance.  It's odd that the test passed locally - but I guess a different version a package may cause minor differences in results beyond the default tolerances in this case.  Glad to get everything passing on Travis CI.

hdoupe:
Sure, I think you can get differences when computers do floating point arithmetic differently or have different levels of floating point precision. I'd guess that's why the results were similar but not exactly the same. I had a class once where people implemented the same algorithm but got wildly different results because the algorithm wasn't stable and we were all on different types of computers. I don't think OG-USA will have a problem on that scale, but it'll be important to decide which results need to be exactly the same and which need to be within a hundredth, a thousandth, etc. I'd imagine this is something that we'll figure out over time as some tolerances need to be adjusted when travis tests fail or when others report similar issues.

PR #449, opened 2019-03-16
Interest rate differential for government debt

jdebacker:
This PR adds an interest rate differential between the return on equity (given by the marginal product of capital) and on government debt.  To ensure that households continue to hold both types of assets, we introduce a household portfolio problem.  ~~We model this as a CES function between debt and equity in the household's portfolio of assets.  This represents household preferences over a mix of debt and equity, where that mix responds to the differential returns, but does not result in corner solutions where only the asset with the higher return is held.~~  We have the household hold debt and equity in the same ratio as the ratio of debt to capital demanded by the government and firm's, respectively.


jdebacker:
@rickecon This PR is ready for your review.  All tests pass locally.  A few commits changed starting values in `SS.py`, but in the end I left those as they were in the master branch.  These values won't solve for certain values of the interest rate wedge, but work well with the default values of parameters.

PR #451, opened 2019-04-01
Partial open economy

jdebacker:
This PR implements a partial open economy version of OG-USA.  In this version, foreign investors purchase a fraction of newly issued government debt and invest in private capital.


jdebacker:
All test pass locally.   This PR is ready for review @rickecon.

jdebacker:
In running some reforms, noticing that the resource constraint is not clearing with an open economy (partial or small open) when there is a government interest rate wedge.  I think that means the wrong interest rate is being passed to an aggregates equation.  Will look into this (theory is laid out in the OG-USA doc updates in [this PR](https://github.com/rickecon/OGUSAdoc/pull/8
)).

PR #462, opened 2019-05-24
Variable storing run start time gets re-assigned

hdoupe:
Rename the `start_time` variable that stores the start time for the run to `run_start_time` so that the final time estimate is accurate.


jdebacker:
Thanks for the update @hdoupe!

PR #482, opened 2019-08-18
fFx error in index.rst

duncanhobbs:
@jdebacker 

This fixes a small error in `index.rst`. 

I think it should say "OG-USA" not "Open CGE".


jdebacker:
Thanks for the fix @duncanhobbs !

PR #488, opened 2019-09-04
Clean up creation of initial time paths

jdebacker:
This PR makes use a function in `utils.py` to create a time path for variables of a specific shape.  It adds tests of this function.  

Some clean up and tests of `utils.to_timepath_shape()` were also added.


jdebacker:
@rickecon This branch passes all tests and is ready for your review.

PR #494, opened 2019-09-16
ParamTools

jdebacker:
This PR updates the OG-USA `Specifications()` class to inherit the `ParamTools.parameters()` class.


ogusa/tests/test_parameters.py
```diff
@@ -26,12 +26,36 @@ def test_create_specs_object():
...
+        'frisch': 0.3,
+    }
+    spec.update_specifications(new_spec_dict)
+    assert spec.frisch == 0.3
+    assert len(spec.errors) == 0
+
+
+def test_update_specification_with_json():
+    spec = Specifications()
+    new_spec_json = """
```
jdebacker: @hdoupe This test is not passing because the JSON is not specified properly:
```
if raise_errors and self._errors:
>           raise self.validation_error
E           paramtools.exceptions.ValidationError: {
E               "frisch": [
E                   "Not a valid number: {'value': [{'value': 0.3}]}."
E               ]
E           }

../../ParamTools/paramtools/parameters.py:244: ValidationError
```

What is the proper format to use here?

jdebacker:
@hdoupe When running OG-USA, I've been pickling the `Specifications` object so that I can easily retrieve parameters from each model run.  I'm noticing that when `Specifications` is built off of `paramtools.parameters`, that I cannot pickle the object because of the dependency on `marshmallow`:

```
>           pickle.dump(spec, open(param_dir, "wb"))
E           _pickle.PicklingError: Can't pickle <class 'marshmallow.schema.DefaultsSchema'>: attribute lookup DefaultsSchema on marshmallow.schema failed

execute.py:88: PicklingError
```

Is that right?  Any work around?

ogusa/tests/test_parameters.py
hdoupe: This should work:

```json
    new_spec_json = """
    {
        "frisch": [{"value": 0.3}]
    }
    """
```

Since you're not using labels, you can simplify it to this:

    new_spec_json = """
    {
        "frisch": 0.3
    }
    """

jdebacker: Got it - thanks @hdoupe.

hdoupe:
@jdebacker I found the source of the pickle problem, but I'm not sure how to solve it yet. I opened an issue in Paramtools: https://github.com/PSLmodels/ParamTools/issues/78

I'll try to get a fix up for this today.

hdoupe:
@jdebacker what do you think about using [`cloudpickle`](https://github.com/cloudpipe/cloudpickle) for dumping your Python objects? It should already be installed since dask lists it as a dependency, and it [serializes python objects that Python's `pickle` does not](https://github.com/cloudpipe/cloudpickle#cloudpickle). 

The issue is that ParamTools dynamically creates Marshmallow classes dynamically using the `type` function, and Python's `pickle` only [supports classes that were defined at run-time](https://docs.python.org/3/library/pickle.html#what-can-be-pickled-and-unpickled). Stackoverflow has a ton of questions related to this ([one](https://stackoverflow.com/questions/11658511/pickling-dynamically-generated-classes), [two](https://stackoverflow.com/questions/24364408/pickling-of-dynamic-class-definition), [three](https://stackoverflow.com/questions/46124823/pickling-dynamically-created-types)). [One solution](https://stackoverflow.com/questions/1947904/how-can-i-pickle-a-nested-class-in-python/11493777#11493777) showed how you can define your own method for pickling in this situation, but it didn't work very well, and I'm a little nervous about supporting something like that even if I can get it to work.

jdebacker:
@hdoupe Thanks for looking into the issue of being able to pickle the class object.  I hadn't heard of `cloudpickle` before, but it sounds useful.  This statement gives me pause:

> Using cloudpickle for long-term object storage is not supported and strongly discouraged.

I'll do some more reading...

hdoupe:
Ah, I agree. That statement was added in this PR https://github.com/cloudpipe/cloudpickle/pull/127 in response to https://github.com/cloudpipe/cloudpickle/issues/123. If you specify the default protocol, [`pickle.DEFAULT_PROTOCOL`](https://docs.python.org/3/library/pickle.html#pickle.DEFAULT_PROTOCOL), in [`cloudpickle.dump`](https://github.com/cloudpipe/cloudpickle/blob/master/cloudpickle/cloudpickle.py#L1112-L1128), then it seems like `cloudpickle` will use a safer, more compatible protocol for creating the pickle file. I think things should work like normal once the file is dumped, since cloudpickle just uses the regular Python pickle for loading pickle files.

jdebacker:
@hdoupe Ok - let me switch things to save the parameters class object with `cloudpickle`.  I'm reassured that it only takes `pickle` to open the files and that `cloudpickle` will be installed with `dask`.  Thanks for this suggestion.

hdoupe:
>  I'm reassured that it only takes pickle to open the files and that `cloudpickle` will be installed with `dask`.

That's what I was thinking, too. If there are any shortcomings or you want to drop the use of `cloudpickle` at some point in the future, I'm happy to revisit this issue.

jdebacker:
@rickecon This PR passes all local tests and is ready for your review.

rickecon:
The failing CI checks are all:
```python
E   ModuleNotFoundError: No module named 'ogusa.parametersbase'
```
stemming from the `test*.py` files that cause the `ogusa/__init__.py` file to run, which has the following import.
```python
from ogusa.parametersbase import *
```
It looks like `ogusa/parametersbase.py` was deleted in this PR. So I think the fix is as simple as deleting that import line in the offending `ogusa/__init__.py` file. I just submitted a [PR](https://github.com/jdebacker/dynamic/pull/12) to your open PR with this fix. See if it passes the CI after integrating.

jdebacker:
@rickecon Thanks for that PR.  I removed a few other references to the old `parametersbase.py` from the docs and package building files also.  I think this is now ready.

PR #496, opened 2019-10-05
Compute Studio params

jdebacker:
This PR will do the following to help with the presentation of OG-USA on compute studio:

1. Add section labels to the parameters in `default_parameters.json`
2. Filter out OG-USA parameters that users do not need to adjust on CS (e.g., model solution tolerances, parameters that require inputting a large matrix of data)
3. Add Tax-Calculator parameters to the CS function file (filtering these as well).

PR #497, opened 2019-10-10
Use context manager for opening files

hdoupe:
This PR uses a context manager for opening all files. This is good hygiene in general, but it may help with the "too many open files" issue that @jdebacker  ran into here: https://github.com/PSLmodels/OG-USA/pull/496#issuecomment-540690715


jdebacker:
@hdoupe Thanks for looking at these read/write command with pickle files.

For the read commands, you can just use the [`utils.safe_read_pickle`](https://github.com/PSLmodels/OG-USA/blob/f017f4d4218e15d1731eb0c05ff93044b259a782/ogusa/utils.py#L366) function.  That will save some lines of code throughout this PR.

hdoupe:
No problem, just pushed the changes using `utils.safe_read_pickle`.

PR #507, opened 2019-10-26
Demographics parameter plots

jdebacker:
This PR moves the code for plotting functions out of `demographics.py` and into `parameter_plots.py`.  This is part of the effort to consolidate graphing as outlined in Issue #92.

### DIFF
--- a/ogusa/parameter_plots.py
+++ b/ogusa/parameter_plots.py
@@ -252,7 +252,7 @@ def plot_chi_n(p, include_title=False, path=None):
 
 
 def plot_fert_rates(fert_func, age_midp, totpers, min_yr, max_yr,
-                    fert_data, fert_rates):
+                    fert_data, fert_rates, output_dir=None):
     '''
     Plot fertility rates from the data along with smoothed function to
     use for model fertility rates.
@@ -299,18 +299,18 @@ def plot_fert_rates(fert_func, age_midp, totpers, min_yr, max_yr,
              'Source: National Vital Statistics Reports, ' +
              'Volume 64, Number 1, January 15, 2015.', fontsize=9)
     plt.tight_layout(rect=(0, 0.035, 1, 1))
-    # Create directory if OUTPUT directory does not already exist
-    output_dir = os.path.join(CUR_PATH, 'OUTPUT', 'Demographics')
-    if os.access(output_dir, os.F_OK) is False:
-        os.makedirs(output_dir)
-    output_path = os.path.join(output_dir, 'fert_rates')
-    plt.savefig(output_path)
-    plt.close()
+    # Save or return figure
+    if output_dir:
+        output_path = os.path.join(output_dir, 'fert_rates')
+        plt.savefig(output_path)
+        plt.close()
+    else:
+        return fig
 
 
 def plot_mort_rates_data(totpers, min_yr, max_yr, age_year_all,
                          mort_rates_all, infmort_rate,
-                         mort_rates):
+                         mort_rates, output_dir=None):
     '''
     Plots mortality rates from the model and data.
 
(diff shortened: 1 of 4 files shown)


jdebacker:
An example of the new plot function added in this PR (all other additions to `parameter_plots.py` are just moving graphing out of `demographics.py`):

```
import ogusa.parameter_plots as pp
base_params = pickle.load(open('./run_examples/CBO/OUTPUT_BASELINE/partial/model_params.pkl', 'rb'))
pp.plot_population(base_params, years_to_plot=[2020, 2040, 2060, 'SS'])
```
pop_distribution

jdebacker:
@rickecon This is ready for your review.  Thanks!

ogusa/parameter_plots.py
```diff
@@ -59,23 +61,24 @@ def plot_mort_rates(p, include_title=False, path=None):
...
     if path is None:
         return fig
     else:
         fig_path = os.path.join(path, "mortality_rates")
         plt.savefig(fig_path)
 
 
-def plot_pop_growth(p, start_year=2019, include_title=False,
-                    num_years_to_plot=150, path=None):
+def plot_pop_growth(p, start_year=2019, num_years_to_plot=150,
```
rickecon: @jdebacker . Why switch the order of these two arguments?

rickecon:
@jdebacker . I am not totally sure why the `codecov/patch` test is not passing and what the exact test is here. It looks like it is because the coverage for `ogusa/parameter_plots.py` went down (see [here](https://codecov.io/gh/PSLmodels/OG-USA/compare/73e8f5611ad1a0e8d45745631b3f36009860b66f...a08d6cfc04644c2d946cae62b2afd0fe723332f1/src/ogusa/parameter_plots.py)).

ogusa/parameter_plots.py
jdebacker: I'm trying to keep the kwargs in a similar order across plotting functions and also group them in a way that makes sense (though I understand order doesn't matter when calling) - so I thought args dealing with the years plotted should be together and those with plot formatting/saving together.

jdebacker:
@rickecon The codecov/path is testing for coverage in the changed files (and codecov/project tests for the entire project).  I added a lot of lines to `parameter_plots.py`, but did not add tests for those functions moved over from `demographics.py`.  

I was going to skip these tests, but let me go ahead and add them (my hesitation was with how I might test a function doesn't return anything, but only saves a file to disk).

rickecon:
Where is the `codecov/patch` requirement for testing percentage? Is it anywhere in the `ogusa/tests/` folder? I am happy to go ahead and merge this if you want to wait on writing those tests for `paramater_plots.py`.

However, I found what looks like a good package for testing our image creation functions. [`Pillow`](https://pillow.readthedocs.io/en/stable/) is billed as a "friendly fork" of the old Python Imaging Library (PIL) package. It is worth noting that `Pillow` and `PIL` cannot exist in the same environment together. But the `pillow` module has a bunch of great commands for verifying the type of an image file.

jdebacker:
@rickecon I just pushed some commits that add tests for all new plotting functions.  Forcing myself to write these was useful and I could see some benefits to be had by adding the output directory as a kwarg of these functions, which also turned out to make testing easier.

Re:

> Where is the codecov/patch requirement for testing percentage?

I think that since we have turned on `codecov` in `.travis.yml`, both the project level and patch level tests are run.  The default percentage to fail the test is that you are not covering less percentage of the new code as you are of the project as a whole.  This [can be turned off](https://docs.codecov.io/docs/commit-status#section-disabling-a-status
) if desired.

rickecon:
@jdebacker. Thanks for pointing out the `.travis.yml`. But where are the criteria set of coverage percentage at both the project and patch level? It is not in `.travis.yml`. I am trying to figure out what the trigger was for that failure of your previous commits and where that is set and adjusted.

jdebacker:
@rickecon The `codecov` tests fail with the following criteria:
* `codecov/patch`: if the coverage of the lines affected in the commit/PR are covered by a lower percentage than the project as a whole.
* `codecov/project`: if the coverage percentage for the project as a whole declines as a result of the commit/PR.

jdebacker:
I should also add that [you can adjust these default criteria](https://docs.codecov.io/docs/commit-status#section-threshold
), but they make sense as a threshold.

rickecon:
@jdebacker. I agree with those criteria. I just did not know what they were and where they were set. Thanks.

PR #508, opened 2019-10-30
Tax plots

jdebacker:
This PR moves the plotting of tax functions out of `txfunc.py` and into `parameter_plots.py`, in response to Issue #92.  

At some point, we should probably change how these functions are used.  Rather than being called within a tax function estimation run, it might be better if they could be called with data saved from the estimation, so that one could used these functions with any cached data (i.e., not have to run the estimation again to plot the tax functions with the data).  Or perhaps they could be structured in a way so as to handle both cases.

In any case this PR will not only move the location of the functions to make `txfunc.py` more clear.  It will not restructure them to work easily with cached data.

### DIFF
--- a/ogusa/parameter_plots.py
+++ b/ogusa/parameter_plots.py
@@ -1,6 +1,7 @@
 import numpy as np
 import os
 import matplotlib.pyplot as plt
+import matplotlib
 from ogusa.constants import GROUP_LABELS
 CUR_PATH = os.path.split(os.path.abspath(__file__))[0]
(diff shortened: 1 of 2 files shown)


jdebacker:
@rickecon This one is ready for your review.  Note that I do not plan on writing tests of the graphing functions moved to `parameter_plots.py` in this PR.  This means that the codecov/patch test is not going to pass.  As I note in the first comment, I think tat in the future, these graphing functions should be rewritten to work with the objects saved after an OG-USA run (e.g., the cached micro data and parameters).  That will be done in a future PR.

Issue #514, opened 2019-11-08
Get data files from web


rickecon:
This issue is a proposed solution to the goal of not storing data in the OG-USA repository. Much of the data we want to use for calibration is already accessible on the web. PR #511  adds the [`wealth.py`](https://github.com/PSLmodels/OG-USA/blob/master/ogusa/wealth.py) module. This module uses Survey of Consumer Finances data in lines 21-22. We want to avoid loading that data in the repository.

This issue proposes to write code using the `requests` and `urllib.request` packages to give the user an option to have the data file locally or to get it from the SCF website. Examples of what I have in mind were written by @hayleefay in the [`hrs_by_age.py`](https://github.com/OpenSourceEcon/CPS_hrs_age/blob/master/hrs_by_age.py) file of the [`CPS_hrs_age`](https://github.com/OpenSourceEcon/CPS_hrs_age) module.

@jdebacker

jdebacker:
Yes - can you can pass a url directly to the `Pandas.read_csv` (and `read_stata`, `read_excel`, etc) function as well.

rickecon:
Many data files that we need to retrieve are zipped. The code updates in PR #517 deal with retrieving zipped files.

PR #517, opened 2019-11-11
Add online data file retrieval functionality to wealth.py

rickecon:
This PR adds online file data retrieval functionality to `wealth.py`, following some of the suggestions in Issue #514. This approach could then be implemented for any large data calls required in OG-USA rather than posting data to this repository. @jdebacker


rickecon:
@jdebacker. What is the URL of the SCF data files called in [`wealth.py`](https://github.com/PSLmodels/OG-USA/blob/master/ogusa/wealth.py) lines 21-23? I am looking at the [SCF data page](https://www.federalreserve.gov/econres/scfindex.htm), and I want to know if the data are the Stata .dta files of the "full data set" or "tables based in intenral and external data" or "summary extract public data".

jdebacker:
@rickecon I'm not sure which were used.

BTW, I see that only `pandas.read_csv()` allows you to read a zip file directly from a URL.  `read_stata()` etc can take a URL, but since the SCF files are zipped, you need to use `requests` or similar to download the zip before you can use `ZipFile` or similar to unzip and then pass to the `pandas.read_x` function.

rickecon:
@jdebacker . Yes. I found the files on the SCF website, and they have nice, uniform URLs. And I have some good code for unzipping them. The first seven packages that I import in `utils.py` are the following.
```python
import os
import sys
import requests
from zipfile import ZipFile
import urllib.request
from tempfile import NamedTemporaryFile
from io import BytesIO StringIO
```
I am trying to write this code so that we can use these new functions to go get any files we want that are stored online elsewhere.

rickecon:
OK, Travis CI is passing. I now just need to write some tests so that `codecov` will pass.

rickecon:
@jdebacker. I have been working with the `wealth.py` and `test_wealth.py` files in this PR. The code for fetching the SCF .zip files is working well. The moments are slightly different that what is currently in the `test_compute_wealth_moments()` test in `test_wealth.py`. The current expected moments are the following.
```python
expected_moments = np.array([
    -0.004314198, 0.019807056, 0.062817143, 0.0644862, 0.121936608,
    0.39036657, 0.344894446, 0.837576949, 4.76976036])
```
The new computed moments that I get from downloading the data and rerunning the test are the following:
```python
expected_moments = np.array([
    -4.31399943e-03, 1.98061283e-02, 6.28176665e-02, 6.44863983e-02,
    1.21936999e-01, 3.90366793e-01, 3.44893838e-01, 8.37576804e-01,
    4.76978025e+00])
```
As you can see, these are really close. I am proposing in this PR to replace these new expected moments with the old ones. And I might actually try to include the 2016 SCF data as well. I am also proposing changing the tolerance in the test's `np.allclose()` command to be a relative 0.001.
```python
assert(np.allclose(expected_moments, test_moments, rtol=0.001))
```

rickecon:
@jdebacker. One last thing before this is ready for your review. The `get_wealth_data()` and `compute_wealth_moments()` functions in `wealth.py` use multiple years of the SCF. The current default is 2013, 2010, and 2007. I noticed that we don't inflation adjust those numbers anywhere in those functions so that they are comparable and rankable. The U.S. CPI for all items the years 2016, 2013, 2010, and 2007 is the following ([link to data](https://fred.stlouisfed.org/series/CPALTT01USA661S)).

| Year  | CPI (2015=100) |
| ----- | -------------: |
| 2016 | 101.26 |
| 2013 | 98.29 |
| 2010 | 92.00 |
| 2007 | 87.48 |

That is a 15% difference between the 2007 numbers and the 2016 numbers. I think we need to inflation adjust. I am going to add the inflation adjustment to this PR.

rickecon:
With the inflation adjustment, the wealth moments change only slightly to the following:
```python
expected_moments = np.array([
    -4.39579116e-03, 1.98789938e-02, 6.28189159e-02, 6.43303277e-02,
    1.21509918e-01, 3.91059782e-01, 3.44792006e-01, 8.37800157e-01,
    4.76076018e+00])
```
If I add the 2016 SCF to the `scf_yrs_list` argument in `get_wealth_data()`, I get the following moment list.
```python
expected_moments = np.array([
        -4.36938131e-03,  1.87063661e-02,  5.89720538e-02,  6.10665862e-02,
        1.17776715e-01,  3.87790368e-01,  3.60041151e-01,  8.45051216e-01,
        4.97530422e+00])
```
I am adding 2016 to the default list. Inflation adjustment is done.

@jdebacker. I think this PR is ready for your review.

ogusa/utils.py
```diff
@@ -701,3 +706,128 @@ def read_cbo_forecast():
...
+        # url = requests.get(file_url) (if using reuests package)
+        url = urllib.request.urlopen(file_url)
+
+        f = NamedTemporaryFile(delete=False)
+        path = f.name
+
+        # url.content (if using requests package)
+        with ZipFile(BytesIO(url.read())) as zipped_file:
+            for contained_file in zipped_file.namelist():
+                for line in zipped_file.open(contained_file).readlines():
```
jdebacker: Can you just do a `.read()` instead of `.readlines()` to avoid this loop?

jdebacker:
@rickecon This is looking good.  The inflation adjustment was a nice catch.

ogusa/utils.py
rickecon: @jdebacker . Not sure. I will test `.read()` here. I am not sure what the `.readlines()` loop line was doing in my [`hrs_by_age.py`](https://github.com/OpenSourceEcon/CPS_hrs_age/blob/master/hrs_by_age.py) code.

rickecon: @jdebacker . I took this loop out and just used the `.read()` method and it seems to pass all the tests locally on my machine, including the correct moments. If this passes the CI and codecov tests on this repo, then I think it is ready for your review and merge.

ogusa/wealth.py
```diff
@@ -1,37 +1,92 @@
...
+        scf_yrs_list (list of 4-digit integers): list of SCF years to
+            import. Currently the largest set of years that will work is
+            [2016, 2013, 2010, 2007]
+        web (Boolean): =True if function retrieves data from internet
+        directory (string or None): local directory location if data are
+            stored on local drive, not use internet (web=False)
+
 
     Returns:
         scf (Pandas DataFrame): pooled cross-sectional data from SCFs
```
jdebacker: You changed this to `df_scf` now

ogusa/wealth.py
```diff
@@ -1,37 +1,92 @@
...
+                    directory=None):
     '''
-    Reads wealth data from the 2007, 2010, and 2013 Survey of Consumer
-    Finances (SCF) files.
+    Reads wealth data from the 2007, 2010, 2013, and 2016 Survey of
+    Consumer Finances (SCF) files.
 
     Args:
-        None
+        scf_yrs_list (list of 4-digit integers): list of SCF years to
```
jdebacker: I think just make this type `(list)` for documentation purposes (so consistent with elsewhere). The part after the colon can describe the need for four digit integers if that's not obvious from the other description.

jdebacker:
@rickecon Had a couple small comments.  But I think once tests added for those other two functions added to `utils.py`, this will be ready to merge.

rickecon:
@jdebacker. I think this PR is ready for you.
* I fixed the small changes in the code you recommended.
* I added two more tests for the two other functions I added in `utils.py`.
* I merged in the changes from your plotting PR #519 that was merged in tonight.
* All Travis CI and codecov tests are passing.

environment.yml
```diff
@@ -18,6 +18,7 @@ dependencies:
 - pycodestyle
 - pylint
 - coverage
+- requests
```
jdebacker: I think we need to add `ZipFile` here.  And maybe some other of the newly imported packages to `utils.py` (I'm not sure which of those are already dependencies of other packages in this environment, but I'm pretty sure that `ZipFile` is not).

jdebacker:
Not super important, but if you want to be sure that the documentation that will be posted to [the OG-USA ReadTheDocs](https://og-usa.readthedocs.io/en/latest/) will reflect your new additions, you should edit `docs/source/utils.rst` to add the three new functions  to the `automodule` list as below:

```
.. _utils:

Utilities Functions
=================================================

**Utilities Functions**

ogusa.utils
------------------------------------------

.. currentmodule:: ogusa.utils

.. automodule:: ogusa.utils
  :members: mkdirs, pct_diff_func, convex_combo, read_file, pickle_file_compare,
    comp_array, comp_scalar, dict_compare, to_timepath_shape, get_initial_path,
    safe_read_pickle, save_return_table, print_progress, fetch_files_from_web, not_connected
```

environment.yml
hdoupe: These packages are in Python's standard library and are installed by default. They shouldn't need to be in the `environment.yaml` file.

rickecon:
In my last commit, I added 8 packages to `environment.yml`, all of which are imported in `utils.py`.
```python
- sys
- os
- urllib.request
- tempfile
- zipfile
- pkg_resources
- io
- requests
```
This creates an error when building the conda environment `conda env update -f environment.yml`.
```python
ResolvePackageNotFound: 
  - urllib.request
  - io
  - pkg_resources
  - zipfile
  - sys
  - tempfile
  - os
```
This happened before when I added these. This time I tried adding these same package to the `conda.recipe/meta.yaml` and got the same error. I think these packages are already dependencies of the packages in the `environment.yml`. I propose that we drop them. @jdebacker

hdoupe:
@rickecon Those packages are in the Python standard library. So, they are already included when you install Python.

rickecon:
@jdebacker I added the new functions in `utils.py` to the `docs/source/utils.rst` file. It now reads:
```rst
:members: mkdirs, pct_diff_func, convex_combo, read_file, pickle_file_compare,
    comp_array, comp_scalar, dict_compare, to_timepath_shape, get_initial_path,
    safe_read_pickle, rate_conversion, save_return_table, Inequality,
    read_cbo_forecast, print_progress, fetch_files_from_web, not_connected
```
I added the three new functions from this PR (`print_progress`, `fetch_files_from_web`, `not_connected`), and I added a function (`rate_conversion`) and a class (`Inequality`) that were already in `utils.py` but were not in `utils.rst`.

It seems to me like the `Inequality` class should be in `wealth.py` rather than `utils.py`.

jdebacker:
@rickecon sounds good.  Wasn't sure if you those were standard packages.  Thanks @hdoupe for the comments.

jdebacker:
@rickecon I don't think it makes sense to have `Inequality` in `wealth.py` since its used for measures on inequality for variables besides wealth.

environment.yml
rickecon: Thanks @hdoupe.

docs/source/utils.rst
```diff
@@ -13,4 +13,5 @@ ogusa.utils
 .. automodule:: ogusa.utils
   :members: mkdirs, pct_diff_func, convex_combo, read_file, pickle_file_compare,
     comp_array, comp_scalar, dict_compare, to_timepath_shape, get_initial_path,
-    safe_read_pickle, save_return_table
+    safe_read_pickle, rate_conversion, save_return_table, Inequality,
```
jdebacker: Ah, I had forgotten to add `Inequality` when it was created.  The way to add the inequality class would be to add

```
.. autoclass:: Inequality
  :members: gini, var_of_logs, ratio_pct1_pct2, top_share
```

This should be at the same level of indentation as `automodule`.

hdoupe:
No problem. Sorry for leaving the same comment twice! I wasn't sure if my first one was buried in the file changes.

jdebacker:
Great additions here @rickecon !  and thanks for cleaning up some stuff too.  Merging.

PR #519, opened 2019-11-13
Add function to plot lifecycle profiles over time path

jdebacker:
This PR adds `output_plots.tpi_profiles()`, which allows the user to plot lifecycle profiles from set of years over the time path.


rickecon:
Looks good to me. @jdebacker

PR #521, opened 2019-11-13
Add better outputs to CS results

jdebacker:
This PR updates the `cs_config/functions.py` module to provide more informative output for the SS results displayed on Compute Studio.

### DIFF
--- a/cs-config/cs_config/functions.py
+++ b/cs-config/cs_config/functions.py
@@ -194,6 +194,8 @@ def comp_output(base_ss, base_params, reform_ss, reform_params,
     '''
     table_title = 'Percentage Changes in Economic Aggregates Between'
     table_title += ' Baseline and Reform Policy'
+    plot_title = 'Percentage Changes in Consumption Between Baseline'
+    plot_title += ' and Reform Policy'
     out_table = ot.macro_table_SS(
         base_ss, reform_ss,
         var_list=['Yss', 'Css', 'Iss_total', 'Gss', 'total_revenue_ss',
@@ -204,8 +206,7 @@ def comp_output(base_ss, base_params, reform_ss, reform_params,
                   'Lss', 'rss', 'wss'], table_format='html')
     fig = op.ability_bar_ss(
         base_ss, base_params, reform_ss, reform_params, var=var,
-        plot_title='Percentage Changes in Conumption Between Baseline' +
-        ' and Reform Policy')
+        plot_title=plot_title)
     in_memory_file = io.BytesIO()
     fig.savefig(in_memory_file, format="png")
     in_memory_file.seek(0)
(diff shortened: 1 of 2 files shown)


rickecon:
@jdebacker . I don't have a good way to check these tables and figures and their output on C/S. Let me know when you think this is ready, and I'll do my best to proof the code.

cs-config/cs_config/functions.py
```diff
@@ -187,14 +188,24 @@ def run_model(meta_param_dict, adjustment):
...
+        base_ss, reform_ss,
+        var_list=['Yss', 'Css', 'Iss_total', 'Gss', 'total_revenue_ss',
+                  'Lss', 'rss', 'wss'], table_format='csv')
+    html_table = ot.macro_table_SS(
+        base_ss, reform_ss,
+        var_list=['Yss', 'Css', 'Iss_total', 'Gss', 'total_revenue_ss',
+                  'Lss', 'rss', 'wss'], table_format='html')
+    fig = op.ability_bar_ss(
+        base_ss, base_params, reform_ss, reform_params, var=var,
+        plot_title='Percentage Changes in Conumption Between Baseline' +
```
rickecon: @jdebacker . This `plot_title` should have an "s" in "...Consumption...".

cs-config/cs_config/functions.py
```diff
@@ -204,9 +215,20 @@ def comp_output(base_ss, base_params, reform_ss, reform_params,
               "media_type": "PNG",
               "title": 'Labor Supply in Baseline and Reform Policy',
```
rickecon: @jdebacker . Shouldn't this "title" be `'Consumption in Baseline and Reform Policy'`? And why not have two figures, one for consumption and one for labor supply? It seems like this PR is removing the labor supply picture and substituting it with consumption.

jdebacker:
@rickecon Thanks for finding those typos.  

You are right - no way to see how these work on CS until merge to master and app updated on CS.  Probably will need to iterate on the table formatting as I think I need to add some styling to the basic `DataFrame.to_html()` output...

rickecon:
@jdebacker . I will merge this as soon as the codecov/patch test passes. It seems like we should be writing tests for the C/S python functions, even if those tests are somewhat trivial.

jdebacker:
@hdoupe Can you advise re coverage of `functions.py` and `helpers.py` for C/S?  There is the C/S testing kit that seems to cover everything, but it's not reflected in the code coverage.  Do you suggest writing other tests to help code coverage numbers or just ignore?

rickecon:
@jdebacker and @hdoupe . If C/S has testing for the python scripts and modules in the `cs-config` folder, then I think we should exclude that folder from `codecov` in this repository.

hdoupe:
I think you're OK excluding `cs-config` from test coverage. The `cs-kit` tests ensure that the functions will work with Compute Studio. Most of the complex logic that needs more detailed testing is offloaded to either ParamTools or OG-USA and is tested thoroughly in each of those projects. If you feel like there is something that isn't tested well enough, you're welcome to write tests in `cs-config/cs_config/tests/`, and they'll be run when the docker image for OG-USA is built.

PR #525, opened 2019-11-28
Remove graphs

jdebacker:
This PR removes the final bit of plotting done outside the `parameter_plots.py` and `output_plots.py` scripts.


jdebacker:
The codecov/patch tests aren't passing because not all conditions in the plotting function are tested.  Tests for each of those conditions will come in a future PR.

This PR is ready for your review, @rickecon.

Issue #502, opened 2019-10-15
Filter behavioral parameters for Compute Studio


jdebacker:
the `functions.py` module should filter the behavioral parameter changes from Compute Studio and these should after the baseline and reform runs.  The policy parameter changes should continue to only affect the reform run.

jdebacker:
Addressed with PR #527

PR #527, opened 2019-12-03
Filter params for OG-USA on CS

jdebacker:
This PR filters out OG-USA parameters that should remain constant across baseline and reform runs on [Compute Studio](https://www.compute.studio), per Issue #502.

### DIFF
--- a/cs-config/cs_config/functions.py
+++ b/cs-config/cs_config/functions.py
@@ -69,7 +69,7 @@ def get_inputs(meta_param_dict):
     filter_list = [
         'chi_n_80', 'chi_b', 'eta', 'zeta', 'constant_demographics',
         'ltilde', 'use_zeta', 'constant_rates', 'zero_taxes',
-        'analytical_mtrs', 'age_specific']
+        'analytical_mtrs', 'age_specific', 'gamma_s', 'epsilon_s']
     for k, v in ogusa_params.dump().items():
         if ((k not in filter_list) and
             (v.get("section_1", False) != "Model Solution Parameters")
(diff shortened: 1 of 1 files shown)


rickecon:
@jdebacker . Because the file changed is `cs-config/cs_config/functions.py` and this pertains to the executing of OG-USA on `compute.studio` only, I shouldn't worry about this too much. But I don't know the context of these changes and what they are doing.

* Where do these filtered lists show up in the code?
* How are they different from standard OG-USA runs?
* Does this change deliver a computational speedup?
* Does it reduce memory usage in an important way?
* If it is valuable, why not incorporate it into the main OG-USA code?

jdebacker:
@rickecon, good questions, see answers below.

> Where do these filtered lists show up in the code?

The list of parameters that are to remain constant between the baseline and reform runs are added with this PR and appear in lines 151-155 of `functions.py`.

> How are they different from standard OG-USA runs?

On CS, the default is that the baseline runs are made with the model's default parameters.  The parameters changed in the web app only affect the reform runs.  What this PR does is to keep certain parameters constant between the baseline and reform runs.  An example of where this might be important: suppose a user would like to simulate a TCJA repeal, but does not like the default value of the Frisch elasticity.  The user can change the Frisch on Compute Studio, but prior to this PR, that would mean that the baseline run would use the current law policy plus the *default Frisch elasticity*, which the reform run would use the new policy *and* the new Frisch elasticity.  What this PR does is ensure that when that user updates a behavioral parameter like the Frisch (or other parameter that we'd like to hold constant across runs -- see my choice of these in the lines referenced above) that updated parameter value is applied to the baseline and reform runs.

> Does this change deliver a computational speedup?

No change to computational speed.

> Does it reduce memory usage in an important way?

No change to memory usage.

> If it is valuable, why not incorporate it into the main OG-USA code?

This is unnecessary when OG-USA is run from source since one has full control over the parameter values or all runs (whereas the current setup on compute studio is to just enter the changes in parameters for the reform run).

rickecon:
@jdebacker . OK, the code looks good to me. Merging.

PR #528, opened 2019-12-10
Sigma values

jdebacker:
With the default starting values, the SS solutions fails to be obtained with values of sigma far from the default of 1.5.

This PR sets up tests for a range of sigma values and sets the bounds in `default_parameters.json` to ensure that users enter a sigma in this range.

### DIFF
--- a/ogusa/tests/test_user_inputs.py
+++ b/ogusa/tests/test_user_inputs.py
@@ -2,7 +2,7 @@ import pytest
 import os
 from ogusa.execute import runner
 CUR_PATH = os.path.abspath(os.path.dirname(__file__))
-PUF_PATH = os.path.join(CUR_PATH, '../puf.csv')
+PUF_PATH = os.path.join(CUR_PATH, '..', 'puf.csv')
 
 
 @pytest.mark.full_run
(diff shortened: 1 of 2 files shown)


jdebacker:
@hdoupe Any suggestions on this code coverage fail?  Changes to `.py` files here only *added* tests - how can coverage decline?

ogusa/tests/test_user_inputs.py
```diff
@@ -2,7 +2,7 @@
 import os
 from ogusa.execute import runner
 CUR_PATH = os.path.abspath(os.path.dirname(__file__))
-PUF_PATH = os.path.join(CUR_PATH, '../puf.csv')
+PUF_PATH = os.path.join(CUR_PATH, '..', 'puf.csv')
```
hdoupe: @jdebacker perhaps, this is the change that's throwing off the code coverage reporting?

hdoupe:
> @hdoupe Any suggestions on this code coverage fail? Changes to `.py` files here only _added_ tests - how can coverage decline?

I'm not sure. I left a comment where I thought the problem might be, but in my opinion, code coverage is a good proxy for how well tested a project is but the change-to-change diffs seem to have a good bit of noise.

jdebacker:
@hdoupe Thanks for the reply.  I think I will ignore the failing codecov reports here.

@rickecon This is ready for your review.

rickecon:
LGTM. Thanks for testing those `sigma` parameter input values and changing the bounds. Merging.

PR #537, opened 2020-02-28
Update cs-functions package to work on compute-studio celery worker

hdoupe:
- Sets processes to false and limits to one thread per worker
- Closes dask client to fix a memory leak that grows over repeated runs


jdebacker:
@hdoupe Thanks for these edits to run with Celery on C/S.

hdoupe:
This is on C/S now. I just submitted 10 simulations. Let's see if everything works.... https://compute.studio/hdoupe/

Issue #372, opened 2018-04-12
Starting values for the model SS


jdebacker:
We have hardcoded starting values for the SS solution to the model.  These should be updated keeping in mind two objectives:

1) A quick solution to the model
2) Robustness to variations in parameters, especially the Frisch elasticity parameter.

jdebacker:
Notes on this after playing with starting values and model parameterizations:
* SS solution can be quite sensitive to initial guess of `r`.  E.g., with initial guess of `r=0.06`, model will not solve for `g_y<0.001`, but with guess of `r=0.05` it will solve for `g_y=0.0`.

jdebacker:
PR #542 made some updates to the initial guesses that seem to be a bit more robust.  But we should continue to think about an algorithm to test some alternative starting values, at least over a small state space.

jdebacker:
PR #662 allows initial guesses for the SS to be passed as arguments from the top-level OG-USA API.

jdebacker:
With PR #662, closing this for now. We can reopen if we feel the need for a more robust solution.

PR #542, opened 2020-03-27
Test SS

jdebacker:
This PR updates the `test_SS.py` test module to cover almost all cases found in `SS.py`.  It also updates and streamlines the results by having tests make comparisons to cached output from the most recent version of OG-USA.

There are still 3 combinations of model specifications that are not tested in the `test_SS.py::test_run_SS` at the moment. Those will be addressed later after better robustness to starting values is built in (e.g., as Issue #372 notes needs to be done).


jdebacker:
@rickecon This PR adds tests of almost all combinations of for the `SS.py` functions.  I few tests are commented out that I could not get to pass.  One of these, a parameterization of `test_SS.test_SS_solver.py` passes locally, but not on Travis CI and I could not determine the reason so finally gave up and commented it out.  I've spent enough time on these tests for now, so will return to the remaining three parameterization that are currently commented out in a future PR.

In getting these tests to pass, I noticed that it's helpful to ensure that the `zeta` parameter sums to exactly one.  I also adjusted the starting values for the steady state so that they work with more combinations of the SS parameterizations tested.

This PR is ready for your review.

PR #543, opened 2020-04-02
Supress warnings

jdebacker:
Root finder operations often raise `RuntimeWarings`, especially when poor initial guesses are used.

This PR adds a constant to `constants.py` that is a flag to show these warnings.  The default is to *not* show RuntimeWarnings, but this can be changed by changing this constant when running the model.

While doing this, I also moved some constants that remained in `utils.py` over to `constants.py` and removed two that were unused.


rickecon:
Looks great. Thanks @jdebacker.

Issue #524, opened 2019-11-24
Precompute and store txfunc .pkl files


rickecon:
Because estimating the tax function parameters from the microsimulation model in `txfunc.py` takes a significant amount of computation time, this issue proposes to precompute the baseline estimated tax functions for various start years and for both the proprietary IRS Public Use File (PUF) and for the publicly available with OG-USA Current Population Survey (CPS) file. These `.pkl` files take very little hard drive space, and each represents at least an hour of computation time.

I propose that we store these files in a new `OG-USA/ogusa/data/txfuncs/` directory. @jdebacker

jdebacker:
Relatedly, besides caching some precomputed tax functions in the repo, let's also allow the user to pass a path to where the tax functions they want to use are stored, if `run_micro=False`.

jdebacker:
These precomputed tax functions were added in PR #544.

PR #544, opened 2020-04-06
Precomputed tax functions

jdebacker:
This PR addresses Issue #524 by adding precomputed tax function parameter estimates to `OG-USA/ogusa/data/tax_functions`.

The PR further enhances the `Specifications` class to allow the user to input a path to cached tax function parameter estimates that s/he may like to use.  The cached files are checked to determine whether they are consistent with the parameters in the current instance of the `Specifications` class.


ogusa/parameters.py
```diff
@@ -484,11 +484,14 @@ def read_tax_func_estimate(self, tax_func_path):
...
-                      'esimtates')
-                flag = 1
+            try:
+                if self.tax_func_type != dict_params['tax_func_type']:
+                    print('Model tax function type is not ' +
+                          'consistent with tax function parameter ' +
+                          'esimtates')
+                    flag = 1
+            except KeyError:
+                pass
```
hdoupe, line 494: @jdebacker I just happened to look over this when I saw you were updating `parameters.py` and have two small suggestions:
- typo in estimates
- it might be more straghtforward to do:
    ```
    if self.taxfunc_type != dict_params.get('tax_func_type', None):
    ...
    ```
    to handle the case where 'tax_func_type' isn't in `dict_params`

jdebacker: Thanks for the suggestions!  Will do.

jdebacker: Actually, @hdoupe while your suggestion avoids the `KeyError` for missing items in the dict, I wanted behavior that was compatible with older version of the tax parameter estimates so that if the key was missing, a flag *is not* raised.  Therefore, I've switched back to the `try: except:`   phrases.

hdoupe: Aha! Sorry for leading you astray. Looks good to me.

jdebacker:
@rickecon This PR is ready for review.  Thanks!

PR #546, opened 2020-04-15
Set version to current

hdoupe:
@jdebacker C/S caches the inputs for a model with its version and meta parameters. If the version isn't updated, then stale updates will be used.


jdebacker:
Thanks for catching this, @hdoupe

hdoupe:
Yep! I'll write back when this is live

hdoupe:
Update is live. Sorry for the delay. Had to push a C/S bug patch real quick.

PR #563, opened 2020-04-29
Fiscal updates

jdebacker:
This PR will make some updates to how fiscal related variables are computed in OG-USA.  The following changes will be made:

1) Fix an indexing error in the calculation of foreign debt holdings, `D_f`.
2) Move the lines of code to determine the path of foreign and domestic debt holdings into the `fiscal.get_D_G_path()` function.
3) Create a function to compute transfers within `fiscal.py` (this and the above help to remove conditional statements related to the balanced budget option from the main TPI (and SS) code.
4) Allow non-zero government spending in the balanced budget configuration.


jdebacker:
@rickecon This PR is ready for review.

ogusa/aggregates.py
```diff
@@ -369,3 +369,43 @@ def resource_constraint(Y, C, G, I, K_f, new_borrowing_f,
                 debt_service_f)
 
     return rc_error
+
+
+def get_K_splits(B, K_demand_open, D_d, zeta_K):
+    r'''
+    Returns total domestic capital as well as amounts domestica capital
```
rickecon, line 376: @jdebacker . This is really a nit pick, but I found it/saw it, so here it is. Get rid of extra "a" in "domestica" misspelling.

jdebacker: Ok - just pushed fix to typo in this branch. Thanks for catching that @rickecon.

rickecon: @jdebacker I will merge this as soon as it finishes passing all the checks, which it will in about 20 minutes. Thanks.

rickecon:
@jdebacker . The tests timed out on the Travis CI testing for Python 3.7. The last test to pass in the 3.7 CI unit tests was:
```
ogusa/tests/test_get_micro_data.py::test_get_data[Baseline] Running current law policy baseline
Running current law policy baseline
Year:  2029
Year:  2030
```
Right after that, the CI testing started giving Dask worker memory warnings like the following,
```
distributed.worker - WARNING - Memory use is high but worker has no data to store to disk.  Perhaps some other process is leaking memory?  Process memory: 3.12 GB -- Worker memory limit: 4.18 GB
```
and the Python 3.7 testing stopped with the following message.
```
The job exceeded the maximum time limit for jobs, and has been terminated.
```
This is a `Dask` parallelism issue. As is noted in Issue #562, we should probably calculate the total amount of RAM available to the Travis CI cores, divide that by the number of cores, and then try to stay under 60% of that on each core.

jdebacker:
@rickecon I restarted the test for 3.7.

I'm not too worried about the dask issue on Travis CI.  It's simple enough to try the test again for those instances when a worker runs into an issue.  This is about the 3rd time this has happened to me in the last 2 weeks and I've always has success on the second attempt.

If we are hitting the time limit, we can flag some additional tests to not run on Travis CI.

PR #565, opened 2020-05-09
Updates to plot and table modules

jdebacker:
This PR makes some changes to the plot and table modules to:
* Include variable names in a column (as opposed to index) so that they are not dropped when used in non-dataframe output (as was happening in `parameter_plots.tax_rate_table()` and `output_plots.macro_table()`.
* Change default start year of model in several cases missed previously.
* Add some more plot labels to `constants.py`


jdebacker:
@rickecon This is ready for you to review.  The 3.7 tests passed.  3.8 timed out, but I restarted.  I removed tests for 3.6.

rickecon:
@jdebacker OK. Reviewing now. What can we do about these test timeouts?

ogusa/output_plots.py
```diff
@@ -335,7 +335,6 @@ def ability_bar(base_tpi, base_params, reform_tpi,
     plt.ylabel(r'Percentage Change in ' + VAR_LABELS[var])
     if plot_title:
         plt.title(plot_title, fontsize=15)
-    plt.legend(loc=9, bbox_to_anchor=(0.5, -0.15), ncol=2)
```
rickecon, line 338: @jdebacker Why are you removing this legend in this figure? I don't see it anywhere else.

jdebacker:
@rickecon  We can flag more tests with the `@pytest.mark.full_run` flag so that they aren't run on Travis CI, but that reduces our test coverage on Travis CI.  We can make tests run faster with dask, but we have run into some memory issues (though we could increase the number of workers from 2 for tests that use parallel processes with less data... 

If this times out again, I'll try the increasing workers route (though not sure how much time that'll shave off). I hate to remove any more tests from Travis CI.

ogusa/output_plots.py
jdebacker: There are no legend labels created for this bar plot and they'd be unnecessary anyway.  The line removed was creating some warnings about using a legend without any labels.

rickecon:
@jdebacker Do you know what the time constraints, processor constraints, and memory constraints are for Travis CI?

ogusa/output_plots.py
rickecon: Got it. Thanks.

jdebacker:
@rickecon Time limit is 50 minutes per job (runs for each version of Python are separate jobs, run in parallel).  There are 2 cores, memory varies (4GB+).

rickecon:
@jdebacker . This PR is ready to merge. While I am waiting for the second run of the Python 3.8 tests to run, I looked at the Travis CI documentation page on common build errors and the section on timeouts ([https://docs.travis-ci.com/user/common-build-problems/#my-builds-are-timing-out](https://docs.travis-ci.com/user/common-build-problems/#my-builds-are-timing-out)). There are output timeouts that are caused by no output being produced for 10 minutes. But I think our timeouts are because of a global time limit that is reached. The [https://docs.travis-ci.com/user/customizing-the-build/#build-timeouts](https://docs.travis-ci.com/user/customizing-the-build/#build-timeouts) page states that the global timeout limit on public repositories is 50 minutes. It seems like we should look into running tests concurrently.

rickecon:
@jdebacker . If we are now only testing Python 3.7 and 3.8, it seems like we should update the `ogusa-dev` environment in [`environment.yml`](https://github.com/PSLmodels/OG-USA/blob/master/environment.yml#L6) to be `python>=3.7` rather than just `python`.

PR #571, opened 2020-05-26
Allow more workers on CS

hdoupe:
This just bumps then number of allowed workers from 1 to 5. I've tested this on the C/S dev server with good results:

Screenshot from 2020-05-25 16-09-32


jdebacker:
@hdoupe Thanks for this PR - can't wait to test this out on C/S!

hdoupe:
@jdebacker this is live on https://compute.studio/PSLmodels/OG-USA now if you want to check it out.

Issue #572, opened 2020-05-28
UBI may not be correctly modeled


MaxGhenis:
I tried a simple OG-USA model run on CS (https://compute.studio/PSLmodels/OG-USA/107), setting all UBI parameters to 12000, and exempting it from taxes. These were the only parameters I changed:
![image](https://user-images.githubusercontent.com/6076111/83161657-8fcc1080-a0bd-11ea-88c6-49ba3f23c29c.png)

The results were unexpected, especially debt reduction:
![image](https://user-images.githubusercontent.com/6076111/83161446-50052900-a0bd-11ea-859f-7193643dc91a.png)

Might the UBI only be going into Tax-Calculator for MTRs/ETRs rather than as a government transfer?

jdebacker:
@MaxGhenis Thanks for opening this issue, I have not done any runs with UBI parameters before, so this is very helpful to see.

First, let me point out that UBI is going through TC and therefore entering OG-USA through it's effects on ETRs/MTRs (though, in your example, MTRs should not be affected).

Second, yes, the results on D are surprising given the lower net tax revenues that should results from this program.  I'm also surprised by some of the other macro changes (esp C).  Let me try to simulate this run locally so I have more results to digest.

cc @rickecon

hdoupe:
@MaxGhenis just fyi, your sim is private. Their private by default on c/s.

MaxGhenis:
Thanks @hdoupe, changed to public.

jdebacker:
@MaxGhenis, I see that OG-USA is not handling the UBI correctly.  This is for at least three reasons:
1. UBI is not considered net tax liability from TC, so we missing the effect of the UBI to the net tax liability out of TC (this is the numerator of the ETR calculation).  However, TC does include the UBI in the measure of expanded income (the denominator of the ETR calculation).  This would tend to push ETRs down after implementation of a UBI (though not in the accurate way, where we'd want both the numerator and denominator to move by the amount of the UBI).  See [here](https://github.com/PSLmodels/OG-USA/blob/49f003a74349dd9e0849b33549cb9dba3da8c8e3/ogusa/get_micro_data.py#L178) for the relevant code.
[...]

PR #573, opened 2020-05-29
OG-USA handling of UBI

jdebacker:
This PR updates OG-USA to properly handle the UBI estimates from Tax-Calculator.  Background discussion for this PR can be found in Issue #572.


rickecon:
@jdebacker and @MaxGhenis. The correct way to incorporate UBI from [Tax-Calculator](https://github.com/PSLmodels/Tax-Calculator) into OG-USA is through the effective tax rate function and not through the marginal tax rate functions. The household budget constraint is the following:
HHBC
UBI will enter the OG-USA model exclusively through the tax liability term T_{s,t}, which is the last term in the budget constraint. The net tax liability is defined as an effective tax rate because it is easier to estimate an effective tax rate function `tau` than it is to estimate a net tax liability function `T`. Let `T_0` be the net tax liability without UBI. Then `T_0` is defined as the following,
ETR0
where `tau_0` is the effective tax rate without UBI, `x` is total labor income, `y` is total capital income, and `x + y` is total income. The difficulty is that `x + y` does not include other types of income like UBI. So we need a new effective tax rate function `tau_1` that incorporates UBI so that the net tax liability `T_1`.

The simplest conceptual specification is to have the new net tax liability with UBI `T_1` equal the original tax liability minus the UBI payment.
T1_simple
But it is difficult to get the units right of UBI dollars from the data to model units. Note that the first term on the right-hand-side of the equivalence is just `T_0` the net tax liability without UBI. So what we want to do is specify a new effective tax rate `tau_1` such that multiplying it by model income `x + y` without UBI income gives the same net tax liability `T_1` with UBI specified above.
T1_new
The effective tax rate is defined in line 188 of [`get_micro_data.py`](https://github.com/PSLmodels/OG-USA/blob/master/ogusa/get_micro_data.py#L188) as the following:
```python
'etr': calc1.array('combined') / calc1.array('expanded_income')
```
The only change that needs to be made is to subtract UBI from the numerator, and not the denominator.
```python
'etr': (calc1.array('combined') - UBI) / calc1.array('expanded_income')
```

jdebacker:
I agree with most of what @rickecon  said, however, I do think `expanded_income` needs to be modified in the definition of the ETR to correctly map to the income definitions in OG-USA.

In particular `expanded_income != x+y`, but rather `expanded_income = x+y+UBI+consumption value of other benefits in TC`.

Therefore, for the ETR to correctly reflect the `tau_1` above, we need to defined it as:
```
'etr': (calc1.array('combined') - UBI) / (calc1.array('expanded_income')  - UBI - other benefits from TC)
```

rickecon:
@jdebacker . I think you're right. The denominator in the ETR calculation in `get_micro_data.py` needs to be as similar as possible to our denominator in the OG-USA model, which is only labor income and capital income.

jdebacker:
@rickecon This PR is now ready for review.

PR #575, opened 2020-06-04
Separate computation of tax liability

jdebacker:
This PR renames the `tax.total_taxes()` function to the more appropriate `net_taxes()`.  The PR also creates functions to compute each component of these individual level taxes.  Having separate functions, especially one for payroll and income tax liabilities, will be helpful in more easily comparing changes in tax liability between the microsimulation model and OG-USA.


jdebacker:
@rickecon This is ready for review.  Note that the data files in `tests/test_io_data` changed only because f the change in key names in the SS and TPI output dictionaries.  I first conducted all tests with old data and verified that they passed.

PR #577, opened 2020-06-29
Update market income definition

jdebacker:
This PR updates the market income calculation in `get_micro_data.py` to be consistent with the new definitions of `benefits_value_total` from Tax-Calculator that includes UBI benefits after [PR #2418](https://github.com/PSLmodels/Tax-Calculator/pull/2418).


jdebacker:
@rickecon This is ready for your review.

PR #579, opened 2020-07-05
Single start year input

jdebacker:
This PR removes the `start_year` parameter from the set of OG-USA inputs on Compute Studio.  Instead, users will only need to choose the start year in the set of meta parameters, which will be passed to OG-USA.


rickecon:
@jdebacker . This looks good to me. Is this ready to be merged?

jdebacker:
@rickecon This is ready to go.

PR #582, opened 2020-07-07
Update default start year to 2021

jdebacker:
I'm done with 2020... This PR updates the default start year to be 2021.  Cached tax function estimates are updated to reflect this.


rickecon:
@jdebacker. It looks like some budget windows need to be updated in the [`test_run_TPI()`](https://github.com/PSLmodels/OG-USA/blob/master/ogusa/tests/test_TPI.py#L294) function of `test_TPI.py`. Four tests of that function are failing.

jdebacker:
@rickecon Test parameters and test data have been updated to account for the change in the default start year and in the cached tax function parameter estimates used in some test.  All are passing locally for me now.

ogusa/tests/test_SS.py
```diff
@@ -323,16 +323,17 @@ def test_SS_fsolve(guesses, args, expected):
                        atol=1e-6))
 
 
-param_updates1 = {}
+param_updates1 = {'start_year': 2020}
```
rickecon, line 326: @jdebacker . Why is the `start_year` set to 2020 throughout `text_SS.py`?

jdebacker:
@rickecon The start year is set to 2020 because the prior output that this function is tested against was computed using 2020 as the start year.  Rather than changing the output for the new parameterization (that has 2021 as the default start year), I just updated the parameters to use 2020 as a start year.

You might wonder why the start year matters for the SS results.  Well, differences are small (e.g., on the order of 1e-05), but the start year did seem to affect SS results.  I didn't look into it too much, but I think it has to do with the population parameters.  With a different start year, the underlying population data (from 2013) are grown out to a different year.  When we find the SS pop distribution, I think this results in a (slightly) different adjustment in immigration rates to hit the SS in some reasonable number of periods.

In future work, we should probably (1) update the underlying population data to something more recent, (2) confirm that the start year does affect the SS population parameters, and (3) update the demographics functions to use those more sophisticated ones you and your students have developed.

rickecon:
@jdebacker . Is that the same reason for the `start_year` of 2020 in the `test_TPI.py` file?

jdebacker:
@rickecon Yes, but we'd expect the start year to affect demographics over the transition path (it was more surprising that the SS was affected).

rickecon:
Got it. Looks good. @jdebacker

PR #599, opened 2020-07-15
CS Fix

jdebacker:
This PR fixes variable names passed to output functions for the C/S app. These names were changed in PR #575, but were never tests/updated for the C/S app until now.

### DIFF
--- a/cs-config/cs_config/functions.py
+++ b/cs-config/cs_config/functions.py
@@ -295,7 +295,7 @@ def comp_output(base_params, base_ss, reform_params, reform_ss,
         in_memory_file2.seek(0)
         fig3 = op. plot_gdp_ratio(
             base_tpi, base_params, reform_tpi, reform_params,
-            var_list=['D', 'G', 'total_revenue'],
+            var_list=['D', 'G', 'total_tax_revenue'],
             plot_type='diff', num_years_to_plot=50,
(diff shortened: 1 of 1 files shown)


cs-config/cs_config/functions.py
```diff
@@ -24,7 +24,7 @@
 TCDIR = os.path.dirname(TCPATH)
 with open(os.path.join(TCDIR, "policy_current_law.json"), "r") as f:
     pcl = json.loads(f.read())
-RES = convert_defaults(pcl)
+RES = pcl#convert_defaults(pcl)
```
jdebacker, line 27: @hdoupe To have this run locally, I had to avoid the use the ``convert_defaults`` function here since Tax-Calc now uses paramtools (at least my local version does).  Do I need to keep this in because C/S is using an older TC version?

hdoupe: Ah, I see. C/S is still on TC 2.9.0 which doesn't have ParamTools. If you'd like, you can run OG-USA with TC on the `master` branch. However, I think a major release is coming soon and it may be better to wait on that before swapping.

jdebacker: @hdoupe I just wanted to confirm.  Let me test locally with the TC 2.9.   I'll merge PR after that completes successfully.

PR #603, opened 2020-07-27
Tests of SS and TPI with delta_tau = 0

jdebacker:
This PR adds cases for SS and TPI tests for a parameterization with `delta_tau=0`.  These cases will be important for testing the when the code is refactored for dynamic firms, as one can test that the code with dynamic firms with zero adjustment costs and `delta_tau=0` passes these tests since those models should be equivalent.


jdebacker:
Going ahead and merging this.  The codecov failures are due to some lost coverage in the `test_SS.py` module (I don't even know why scripts in `ogusa/tests` are counted towards code coverage).

PR #615, opened 2020-08-08
Add additional files/directories to gitignore

jdebacker:
This PR adds additional directories to the `.gitignore` file to ignore output generated when running the example script and tests.  A large datafile needed for testing is also ignored.


rickecon:
This looks good to me. Is it ready to go?

jdebacker:
@rickecon both 614 and 615 can be merged.

jdebacker:
@rickecon I added some more files to ignore here.  I am hoping that now all files generated with `pytest` and `run_ogusa_example.py` are ignored.

One file in my repo that I'm hesitating adding or ignoring is an excel workbook that I use to create tests by working through function in a spreadsheet.  I think it's useful to for creating new tests and editing old ones, but it's kind of messy and would be hard to version control, I think.  Let me know your thoughts.

rickecon:
@jdebacker . I think we should ignore the Excel spreadsheet you are referring to. Also, the CI tests for these changes all timed out. I am restarting them one by one to see if we can get them to pass. It looks like everything is passing up to the time out.

rickecon:
@jdebacker . I have restarted both sets of CI tests separately twice and they keep timing out. I am not sure why this is happening.

jdebacker:
It's because I added an additional parameterization to `test_SS.test_SS_solver()` when we were already close to the Travis CI time limit.  With my most recent commit, I moved a parameterization out of `test_SS.test_SS_solver()` to `test_SS.test_SS_solver_extra()`, which contain extra tests that will not run on Travis CI (because of time constraint issues).

rickecon:
OK. Then this looks ready to be merged as soon as it passes CI. Thanks @jdebacker.

rickecon:
@jdebacker . Python 3.8 CI tests all passed, but Python 3.7 tests timed out. I just restarted the Python 3.7 tests.

PR #629, opened 2020-08-24
Function to plot tax functions in 2 dimensions

jdebacker:
This PR creates a new function in `parameters_plots.py` to aid users who wish to visualize tax function in 2 dimensions.  The new function, `plot_2D_taxfunc`, allows the user to enter lists dictionaries of tax function parameter estimates of the form output from OG-USA's tax function estimation, plot the function over labor or capital income, and specify the constant value of the other income source (that not represented on the x-axis). The resulting plot looks like:

Unknown-5


jdebacker:
This PR is still a work in progress.  The following items need to be addressed:
1) Add an optional argument where the user can supply a data frame of data create a scatter plot over which the functions are plotted.
2) Test the newly added `tax.tax_rates()` function.
3) See if something like the `tax.tax_rates()` function can help consolidate code that repeats the same (or similar) calculations of tax rates in `tax.ETR_income`, `tax.MTR_income`, and `txfunc.get_tax_rates()`.

jdebacker:
It took a while to figure out a nice way to compute weighted means by group, but able to get plots with scatter of rates by percentile in the data to show against the functions (not shown here, but you can have multiple data series to plot, e.g., both baseline and reform functions on plot with scatter of baseline and reform data): 

test_funcs

Thanks @MaxGhenis for suggesting this.

MaxGhenis:
Just curious, does this chart reflect an actual fitting from OG-USA?

> test_funcs

jdebacker:
@MaxGhenis It doesn't - I was just testing and those data and the parameter estimates come from different policy.

jdebacker:
This PR is close to finished (need to check code coverage issues).  While the initial intent was to simply add one more plotting utility that proves useful in visualizing and comparing tax functions, this addition lead to the consideration of how/where tax rates were computed in the model.

This PR now consolidates all the calculations of tax rates, which had been done in `txfunc.py` and `tax.py` into one function, `txfunc.get_tax_rates()`.  Because the prior functions had specified different a different ordering of the tax function parameters, their is not a change to how the parameters are stored in arrays in the output of the tax function estimation, this resulted in changes in (1) the cached tax function files in `OG-USA/ogusa/data/tax_functions` and (2) changes in the arrays specified in `test_SS.py`, `test_TPI.py`, `test_aggregates.py`, `test_txfunc.py`.    In these test files, I changed the shape of arrays of tax function parameters in cached test data whenever possible.  For tests that use the cached baseline tax function files in `OG-USA/ogusa/data/tax_functions` (the tests of `TPI.run_TPI`), I had to update the cached output files that new results are tested against.

ogusa/parameter_plots.py
```diff
@@ -834,3 +866,133 @@ def plot_income_data(ages, abil_midp, abil_pcts, emat, output_dir=None,
...
+            data_to_plot = txfunc.tax_data_sample(data_to_plot)
+            # set number of bins to 100 or bins of $1000 dollars
+            n_bins = min(100, np.floor_divide(max_inc_amt, 1000))
+            # need to compute weighted averages by group...
+            wm = lambda x: np.average(
+                x, weights=data_to_plot.loc[x.index, "weight"])
+            data_to_plot['inc_bin'] = pd.cut(data_to_plot[key1], n_bins)
+            groups = pd.DataFrame(data_to_plot.groupby(["inc_bin"]).agg(
+                rate=(rate_type, wm), income=(key1, wm)))
+            plt.scatter(groups['income'], groups['rate'], alpha=0.1)
```
MaxGhenis, line 985: WDYT of adjusting alpha based on the number of tax units in the group? Right now it looks like 1 gets the same visual representation as 1M (is that right?) which makes it harder to assess the fit.

MaxGhenis (commented):
.

ogusa/parameter_plots.py
jdebacker: @MaxGhenis One reason I want to keep alpha low is because on can pass a list of data objects (e.g., one for each of several specifications of policy) and I want one to be able to see those points and the tax functions easily.  Also, I think it could be distracting to have some dots for the same data series that are lighter or darker in color.   

I view these graphs as a way to quickly diagnose issues with shape and fit (e.g., see Issue #635).  If one wants more detail to see what's going on, it's probably best to look at statistical measures of fit, such as the SSE's that are stored in the tax function estimation output.

jdebacker:
@rickecon This PR is ready to review.  The failures with code coverage are coming from coverage of test files, which we had tried to remove from the code coverage results, but apparently haven't done successfully.

rickecon:
@jdebacker . Looks great. Thanks for doing this.

PR #631, opened 2020-08-27
Fix get_tr docstring: returns transfers, not bequests

MaxGhenis:

PR #637, opened 2020-09-02
Updates for C/S with Tax-Calc 3.0.0

jdebacker:
This PR updates the `cs-config` files to work with the Tax-Calculator 3.0.0 and its use of ParamTools.


rickecon:
@hdoupe made the following comments in merged PR #631 that was on a different issue.

> Hey @MaxGhenis and @jdebacker, @Peter-Metz and have started going through the process of updating some of the cs-config code for Tax-Cruncher here: PSLmodels/Tax-Cruncher#38.
> 
> Tax-Cruncher doesn't have the CPS/PUF distinction so it's a little more straightforward. I'm planning to do the update to account for some parameters only being compatible with one of the files tomorrow or Wednesday.

jdebacker:
@rickecon Yes, his updates as of about 10pm last night are included here.  He made a few more changes today that I'll try to incorporate, but all tests are passing for `cs_config` currently.

jdebacker:
@rickecon I've successfully run the full testing suite with the changes in this branch.  It's ready for your review.

cs-config/cs_config/inputs.py
```diff
@@ -0,0 +1,77 @@
...
+
+def convert_policy_defaults(meta_params: Parameters, policy_params: Parameters):
+    """
+    Convert defaults for taxcalc's Policy class to work with C/S.
+    """
+    policy_params.array_first = False
+    policy_params.uses_extend_func = False
+    init = policy_params.dump()
+    defaults = convert_sections(init)
+    defaults = convert_data_source(defaults, meta_params.data_source)
```
hdoupe, line 14: @jdebacker I think you may be a version behind since you're missing this line: https://github.com/PSLmodels/Tax-Cruncher/blob/master/cs-config/cs_config/inputs.py#L15

jdebacker: Ok - just added this line.  Thanks for reviewing the changes @hdoupe!

rickecon:
@jdebacker. @hdoupe's last comment was:
> Sorry @jdebacker you need the function too. You're best bet is to just copy the file over in case I missed something else: https://github.com/PSLmodels/Tax-Cruncher/blob/master/cs-config/cs_config/inputs.py
>
> Definitely looking forward to having a better set up for distributing these functions.

 It looks like this comment is satisfied/resolved by what we currently have in this PR in OG-USA/cs-config/cs_config/inputs.py. If that is correct, then this PR is ready to merge.

jdebacker:
Ok, latest commit updates `inputs.py` to match that used in `Tax-Cruncher`.

Note that OG-USA only uses one function from `inputs.py`, `convert_behavior_adjustment()`, so all tests pass before the last two commits and continue to pass after.

hdoupe:
Thanks for updating @jdebacker and review @rickecon . LGTM

PR #639, opened 2020-09-08
add test_run_example

jpycroft:
This PR is to add test_run_example to ogusa/tests/.

This pytest determines whether run_examples/run_ogusa_example.py runs for at least 5 minutes.

This is done by setting off the "main" function in run_ogusa_example.py, waiting for 300 seconds and checking whether it is still running (and stopping it, if it is).


ogusa/tests/test_run_example.py
```diff
@@ -0,0 +1,43 @@
+#### import modules (same as run_ogusa_example)
+import multiprocessing
+from distributed import Client
```
rickecon, line 3: @jpycroft. I think you can get rid of the following import lines:
```python
...
from distributed import Client
...
import taxcalc
from taxcalc import Calculator
from ogusa import output_tables as ot
from ogusa import output_plots as op
from ogusa.execute import runner
from ogusa.constants import REFORM_DIR, BASELINE_DIR
from ogusa.utils import safe_read_pickle
```

ogusa/tests/test_run_example.py
```diff
@@ -0,0 +1,43 @@
...
+from taxcalc import Calculator
+from ogusa import output_tables as ot
+from ogusa import output_plots as op
+from ogusa.execute import runner
+from ogusa.constants import REFORM_DIR, BASELINE_DIR
+from ogusa.utils import safe_read_pickle
+
+currentdir = os.path.dirname(os.path.realpath(__file__))
+parentdir = os.path.dirname(currentdir)
+sys.path.append(parentdir)
```
rickecon, line 17: @jpycroft  and @jdebacker. We probably need to use a different method of loading this module than the `sys.path.append`. It is discouraged to amend the path like this. I think that every time this is called, this will add a string to the `sys.path` list argument. But I have spent the last three hours trying to get a way and none work yet.

ogusa/tests/test_run_example.py
jpycroft: Thanks Rick. I've checked that it works without these lines, and now removed them. (An earlier solution I was playing needed them, but this one does not.)

ogusa/tests/test_run_example.py
jpycroft: @rickecon and @jdebacker. Point taken Rick. 

Let me share an alternative solution, which might help. The code is here: 
https://github.com/jpycroft/OG-USA/blob/jptest/ogusa/tests/test_run_example-wi-pathlib.py

It does work, though I'd say it's less elegant, as you need to change run_ogusa_example.py.
Using pathlib, I navigate to and execute the whole file "run_ogusa_example.py". 
However, the "if __name__ == "__main__":" condition fails, so you have to add another condition:
    if __name__ == "test_run_example-wi-pathlib":
        main()

I haven't _yet_ found a way of using pathlib to be able to call run_ogusa_example.main() directly. I will keep looking though!

jpycroft: @rickecon and @jdebacker. I've uploaded a new test_run_example file. 

It uses importlib.util in place of sys.path.append to locate the module. It also avoids any need to alter run_ogusa_example.py in any way (see previous comment).

jdebacker:
@jpycroft This PR looks good, except that the new test is failing on Travis CI.  I don't know why, but the assertion is failing (`timetest=False` when you want `timetest=True`).

jpycroft:
@jdebacker. Thanks Jason. I'm not so familiar with Travis CI, but I'll take a look.

rickecon:
The error happening in the continuous integration testing on Travis CI can be observed by clicking the "Details" link next to the red "x" on the Travis CI branch of the PR status box near the bottom of this page. Then in the "Jobs and Stages" section, click on the link for Job 2007.2 associated with Python 3.8. If you scroll down, you'll see the output from all the CI tests. The last one is your new `test_run_example.py` test. Here is a screenshot of the output.

TravisCIoutput

I think this is a problem with `importlib` and `pathlib` not being in the `environment.yml` because the test works when I run it on my own machine. I am testing it out now. I have been reading a good [tutorial on imports](https://realpython.com/python-import/) from Real Python. I think it is open access, but I don't know. I will submit a PR to @jpycroft's branch if my `environment.yml` update fixes the testing.

But again, this test works great locally. So we just now need to make it work on the Travis CI server.

rickecon:
@jpycroft and @jdebacker. I just submitted a [PR](https://github.com/jpycroft/OG-USA/pull/1) to Jon's branch. I think the error might be that `importlib` and `pathlib` libraries need to be part of the `environment.yml` as mentioned in my above comment. We'll see if this fixes the problem.

hdoupe:
@jpycroft Have you tried using a Dask `Future` to run this and handle the timeout? It has a `result` method with a timeout parameter that might be easier to work with: https://docs.dask.org/en/latest/futures.html#submit-tasks

Also @rickecon [`pathlib`](https://docs.python.org/3/library/pathlib.html) and [`importlib`](https://docs.python.org/3/library/importlib.html) are in the standard library and shouldn't need to be installed.

rickecon:
Yes. @hdoupe is right. My changes to `environment.yml` were not the issue. @jpycroft you can remove the `import importlib` and `import pathlib` lines from the `environment.yml` file and then commit those changes.

jpycroft:
@hdoupe and @rickecon.  
Hank, I had not thought to try dask futures for the timeout. I'll look into it. Thanks! 
And I'll also remove pathlib & importlib from environment.yml.

jdebacker:
FYI, @jpycroft to help you debug any Travis CI issues you can look at the log files.  To do this: Click "Details" next to the failed text in "checks have failed" box -> go to "jobs and stages" -> click on a job number -> scroll down to see the traceback of the issue (will look like the traceback from `pytest` run locally.

jpycroft:
@jdebacker and @rickecon. Looking through the Travis CI error report, it said that "Timed out trying to connect to 'tcp://127.0.0.1:34713' after 10 s". I wondered whether this was simply because the test includes a 300-second wait. To test, I replaced with a **5 second** wait, and yes, the Travis CI passed. (but obviously, that's not the test we want.)

After passing Travis CI, code coverage is checked, and codecov says that adding this PR would **reduce** coverage by 0.09%. (It doesn't seem fair to _add_ a test and reduce coverage.)

Perhaps we can speak about it tomorrow at the [developer meeting](https://github.com/PSLmodels/OG-USA/issues/641).

rickecon:
Two comments
1. Regarding the 300 seconds versus 5 seconds wait times in the `test_run_example.py` script, we should see if we can get this to work with the 300 second wait time. If it fails because of a timeout, we can do one of three things.
    * Restart the tests manually and see if they pass.
    * Remove a test from the CI run set of tests so the new `test_run_example.py` doesn't add to the run time.
    * Put `test_run_example.py` in the `full_run` set of tests so it does not run in CI.
2. The coverage failure is probably the addition of the `if __name__ == "test_run_example-wi-pathlib":` lines added at the end of `run_ogusa_example.py`, which adds lines. However, the new `test_run_example.py` should cover these lines. I don't know why the code coverage is failing. The only other thing it could be is the two lines added at the end of `environment.yml`. Either way, I think we should merge this in spite of the coverage fail.

cc: @jpycroft @jdebacker

jpycroft:
@rickecon and @jdebacker. Rick, to your second point, the `if __name__ == "test_run_example-wi-pathlib":` lines are no longer needed. I've now removed them. Thanks for catching this.

rickecon:
In the OG-USA developer meeting today, we decided to do the following to finish this PR.
* Reinstate the 300-second test (done in commit [76416f1](https://github.com/PSLmodels/OG-USA/pull/639/commits/76416f12cc0778d82e31393bcdbab37d8bd7c440))
* @jdebacker / @rickecon will submit a PR to @jpycroft's branch that removes 5 minutes of Travis CI tests so they can run with the `test_run_example.py` test
* @jdebacker / @rickecon will submit a PR to @jpycroft's branch that updates the `codecov.yml` file to ignore files in `ogusa/tests`.

rickecon:
@jdebacker and @jpycroft. I just submitted [a PR](https://github.com/jpycroft/OG-USA/pull/3) to @jpycroft's branch for the current PR to OG-USA master. My PR to @jpycroft's `jptest` branch updates his PR with the following:
* Updates @jpycroft's PR to include recently merged OG-USA PR's. This explains the changes in the `cs-config` folder (2 files), the `docs` folder (7 files), and `environment.yml`,  
* Updates the `codecov.yml` to ignore any files in the `ogusa/tests` folder. This should solve the CodeCov test failures that we are seeing from this PR.
* Updates the set of tests that automatically run on Travis CI. This PR removes 2 tests  from the `test_utils.py` script [`test_pickle_file_compare()` and `test_fetch_files_from_web()`] and 2 tests from the `test_income.py` script [`test_get_e_interp()` and `test_get_e_interp_exception()`]. This should fix the failure of Travis CI due to timeout.

jdebacker:
Ok - I just opened another PR to @jpycroft 's branch that fixes a non-Travis test issue.

> Updates the set of tests that automatically run on Travis CI. This PR removes 2 tests from the test_utils.py script [test_pickle_file_compare() and test_fetch_files_from_web()] and 2 tests from the test_income.py script [test_get_e_interp() and test_get_e_interp_exception()]. This should fix the failure of Travis CI due to timeout.

These are all pretty quick tests so only save a few seconds.   I moved a parameterization of the `test_TPI.py::test_run_TPI` tests into the set that don't run on Travis.   I don't know if this alone buys us enough time, but if it does, we can add back the above tests.

Also, before merging, we need to remove the `@pytest.mark.full_run` flag from the new test and can remove the import of the `sys` module from `run_ogusa_example.py`.

jpycroft:
@rickecon and @jdebacker. (I thought this was done, but ... )
The Travis CI check still fails with this message:
`OSError: Timed out trying to connect to 'tcp://127.0.0.1:43383' after 10 s: Timed out trying to connect to 'tcp://127.0.0.1:43383' after 10 s: connect() didn't finish in time`
I suspect that the timeout error is not a timeout for the whole Travis CI check, but it is specific to test_run_example. (If so, moving other tests to full_run won't help.) 

Instead, I'd suggest adding test_run_example to full_run (Rick's solution 3). I've gone ahead and added:
`@pytest.mark.full_run`
above "def test_run_ogusa_example" in test_run_example.py.

And now Travis CI passes, because it's not checking the new test.

rickecon:
@jpycroft @jdebacker. I think we want to keep `test_run_example.py` in the Travis CI set of tests. Let's remove the `@pytest.mark.full_run` flag from the `test_run_ogusa_example` test and find another solution for getting the Travis CI tests to pass.

I am a little bit puzzled why the fix to `codecov.yml` didn't fix the coverage tests.

jdebacker:
> @jpycroft @jdebacker. I think we want to keep test_run_example.py in the Travis CI set of tests. Let's remove the @pytest.mark.full_run flag from the test_run_ogusa_example test and find another solution for getting the Travis CI tests to pass.

I agree - this should be included in the Travis tests.

> I am a little bit puzzled why the fix to codecov.yml didn't fix the coverage tests.

Me neither.  Would these changes not reflected until this new coverage yml is in the master?

Regarding Travis CI not liking the `sleep`, we might have two options:
1. Make sure we are doing something every few seconds to keep the Travis CI machine awake.  e.g., replace:
```
time.sleep(300)
```

With:
``
for i in range(30):
    time.sleep(10)
    print("Still here!")
```

2. Stop using Travis CI and instead do our tests via a GH action (as @MaxGhenis has done in other repos) (though I'm not sure we wouldn't run into the same issue.

**Comment by jdebacker** [comment]
@rickecon Questions about the above suggestions?

**Comment by rickecon** [comment]
OK. The code coverage failures are second priority. I am happy to merge with those. Let's try @jdebacker's suggestion and replace the `time.sleep(300)` command with the following:
```python
for i in range(30):
    time.sleep(10)
    print("Still here!")
```
If this works, I will not have any explanation for why the `time.sleep(300)` command did not work. But, oh well.

@jpycroft . Here are the steps to do.
1. Merge in @jdebacker's PR to your branch and see if it passes the tests. If so, then we don't need to do the suggested change above.
2. If it doesn't pass Travis CI in step (1), commit a change that substitutes `time.sleep(300)` with the `for` loop of 10 second increments. Let's see if that passes.

jdebacker:
To explain, I am suggesting creating a loop with smaller time intervals rather than one long time interval to handle the failure reported by @jpycroft above:

>  Looking through the Travis CI error report, it said that "Timed out trying to connect to 'tcp://127.0.0.1:34713' after 10 s". I wondered whether this was simply because the test includes a 300-second wait. To test, I replaced with a 5 second wait, and yes, the Travis CI passed. (but obviously, that's not the test we want.)

rickecon:
@jdebacker. I agree with your comment and assessment above. Do you agree with my next steps for @jpycroft to follow?
1. Merge in @jdebacker's PR to your branch and see if it passes the tests. If so, then we don't need to do the suggested change above.
2. If it doesn't pass Travis CI in step (1), commit a change that substitutes `time.sleep(300)` with the for loop of 10 second increments. Let's see if that passes.

jdebacker:
@rickecon, @jpycroft also needs to remove the `@pytest.mark.full_run` marker from the new test.

rickecon:
Ugh. It looks like the `test_run_ogusa_example()` test is returning `timetest == False` on Travis CI, even with the loop of 60 5-second intervals. Let's test this again locally (which will probably work). Then we have to determine why it is not working on Travis CI.

jpycroft:
@rickecon and @jdebacker.
Same error using 30 x 10 seconds wait. 
`OSError: Timed out trying to connect to 'tcp://127.0.0.1:41477' after 10 s: connect() didn't finish in time.`
Try again with 60 x 5 second waits. This is in case in has to be strictly _less than_ 10 seconds ... and still the same error:
`OSError: Timed out trying to connect to 'tcp://127.0.0.1:45059' after 10 s: connect() didn't finish in time`
(Rick, I see you'd already noticed this.)
It does work locally:
========= 1 passed in 301.75s (0:05:01) ====================

jdebacker:
Ok, since you are starting a process, `p`, you might need to check in on that process.  

Maybe:
```
for i in range(60):
         time.sleep(5)
         p.is_alive()
         print("Still here!")
```

jpycroft:
@rickecon and @jdebacker. So unfortunately, the trick of adding p.is_alive() into the loop did not get it to pass the Travis CI tests. (The test does pass locally.) 
The error is still:
`OSError: Timed out trying to connect to 'tcp://127.0.0.1:38571' after 10 s: connect() didn't finish in time`
Perhaps interesting `Still here!` does print out 60 times.

jdebacker:
@jpycroft I don't know if it'll make a difference, but I added a new GH Action that will run the tests (this could substitute for Travis CI).  This *might* solve the issue with the timeout if that timeout is specific to something about Travis CI (which is maybe since things run fine locally).  If you make a trivial commit and push it to this branch, I think that action will be triggered and we can see if the tests pass in that environment.

A suggestion for a trivial commit (that might be informative).  Change the `print('Still here!")` to `print("Still here? ", p.is_alive())`.

jpycroft:
@rickecon and @jdebacker. 
Still not passing, but Jason your idea to print out p.is_alive() maybe is helpful. The print out in Travis CI & GH Action:
```
OSError: Timed out trying to connect to 'tcp://127.0.0.1:41079' after 10 s: Timed out trying to connect to 'tcp://127.0.0.1:41079' after 10 s: connect() didn't finish in time
ogusa/tests/test_run_example.py::test_run_ogusa_example 
Still here! True
Still here! True
Still here! False
Still here! False
Still here! False
Still here! False
Still here! False
...

```
i.e. for 10 seconds the process is alive, but by 15 seconds it's not.

hdoupe:
It looks like this could be an issue with Dask not connecting to a scheduler. Note that the initial `OSError` is thrown by the `distributed/comm/core.py` module.

Screenshot from 2020-10-02 10-17-18

jdebacker:
@hdoupe Any thoughts why this might lose connection on Travis, since local runs have no issue?

Also, we use `dask` in other tests, with  no issue.  Maybe instead of using the `multiprocessing` library:
```
p = multiprocessing.Process(target = f,
                                 name="run_ogusa_example", args=())
```

we use a `dask` client (as in other tests):
```
@pytest.fixture(scope="module")
def dask_client():
    cluster = LocalCluster(n_workers=NUM_WORKERS, threads_per_worker=2)
    client = Client(cluster)
    yield client
    # teardown
    client.close()
    cluster.close()
```

We can then check in on the `TaskState` of that process?

hdoupe:
@jdebacker I agree about using dask. That may make this easier.

I think you could do something like this:

```python
import dask.distributed.TimeoutError

future = client.submit(call_run_ogusa_example)
try:
    # https://distributed.dask.org/en/latest/api.html#distributed.Future.result
    result = future.result(timeout=300)
except dask.distributed.TimeoutError:
    print("Success")

```

This should throw an exception after the timeout period if things are running smoothly. However, if there is another error than the `result()` method will throw it like a normal exception and the test should fail.

jpycroft:
@jdebacker and @rickecon.
I should post an update on what I've done to try out Hank's solution. 

Earlier on my problem was that the test was always passing. So the TimeoutError was never returned (even when I deliberately killed run_ogusa_example). My code is the following:

```
from dask.distributed import Client, TimeoutError
... 
def test_run_ogusa_example():
    timetest = True
    try:
        future = client.submit(call_run_ogusa_example)
        # https://distributed.dask.org/en/latest/api.html#distributed.Future.result
        result = future.result(timeout=30)

    except TimeoutError:
        print("TimeoutError returned")
        timetest = False
    
    assert timetest == True
```
I thought the problem was with Dask restarting the process automatically, and so not timing out. The output was like this:

```
stop to test if test_run_example fails
distributed.nanny - WARNING - Restarting worker
stop to test if test_run_example fails
distributed.nanny - WARNING - Restarting worker
stop to test if test_run_example fails
distributed.nanny - WARNING - Restarting worker
stop to test if test_run_example fails
...
```

Now the problem has flipped. Now I _always_ get a TimeoutError. When it should be passing I get an error that says that the Port is in use:

```
C:\Users\jonat\anaconda3\envs\ogusa-dev\lib\site-packages\distributed\node.py:155: UserWarning: Port 8787 
is already in use.
Perhaps you already have a cluster running?
```
I spent time on StackOverflow trying to investigate, but no solutions yet.

Jason, before trying out your style of solution, I need to stop trying to ad hoc this, and go learn more about Dask!

jdebacker:
@jpycroft Let's do the following an then merge this:
1) Add a `pytest` marker to the new test to only run on  a "full run"
2) Remove the `pytest` markers from `tests_utils.py`
3) Revert the changes made to `run_ogusa_example.py` -- there should be no changes to that script in this PR.

jpycroft:
@jdebacker and @rickecon. 
As agreed: 
1. the pytest full_run decorator has been added to test_run_example.py and 
2. removed test_utils.py entirely from PR (so as not to add the full_run decorator). 

The test_run_example.py code is an early October version. It runs locally for me, passing the test normally and failing when I kill run_ogusa_example.py deliberately. Please check it runs locally for you too. 

A number of changes were added to test_TPI.py and test_income.py, including a full_run decorator to some tests. I've left these in.

The other change is adding the tests folder to be ignored in codecov.yml.

jdebacker:
@jpycroft once the comments above are addressed, I think this is ready to merge.

jpycroft:
@jdebacker and @rickecon. Jason, I've removed the two full_run decorators from test_income.py, which now doesn't change and so is no longer part of the PR.

jdebacker:
Great - thanks @jpycroft!  Merging now.

PR #644, opened 2020-09-18
Clean up equations/math within documentation

chusloj:
This PR addresses the equations/math/`LaTeX` issues within #643. Issues with math not rendering correctly or cleanly were addressed in the following sections:

- Stationarization
- Equilibrium
- Exogeneous parameters
- Demographics
- Lifetime earnings profiles
- Derivations

I've found a couple things for how the Jupyter Book docs process math and text in general:

- JB is very sensitive to tabs and spaces between lines, especially when it comes to inserting math.
- The [MyST guide to markdown](https://jupyterbook.org/reference/cheatsheet.html?highlight=list#lists) should be followed exactly regarding spaces and tabs for creating lists in order for JB to correctly render lists as specified.

PR #646, opened 2020-09-22
Remove old docs files

jdebacker:
This PR removes old docs files associated with the Sphinx docs and the OG-USA guide pdf file which has source file in another repo.  These have recently been replaced with the OG-USA Jupyter-Book documentation.


jdebacker:
At some point, will need to take down the readthedocs page: [https://og-usa.readthedocs.io](https://og-usa.readthedocs.io) as it will no longer update after this PR.  I'm just not sure how to do that... not seeing option to delete a project from my account.

rickecon:
@jdebacker. I would leave [OGUSAdoc.pdf](docs/OGUSAdoc.pdf) in the repository for now. There is still a lot of work to get the Jupyter Book documentation to look like the LaTeX-produced PDF. For example, I used the LaTeX-produced PDF to clean up the `demographics.md` documentation file in PR #644. There is a lot more of that still to do.

Regarding deleting the `readthedocs` site, I found this [StackOverflow link](https://stackoverflow.com/questions/48062993/is-there-a-way-to-delete-a-project-on-readthedocs-org) that seems like it is the answer.

jdebacker:
Ok - I thought things would be in good enough shape after PR #644.  

We can let this sit until we're ready to remove the PDF.

I'll take down the readthedocs when this PR is merged.  Thanks for the link.

rickecon:
@jdebacker. I think we should move ahead with this PR and remove the `readthedocs` files and site. I just think we should keep the `OGUSAdoc.pdf` in place until the Jupyter Book documentation is closer. Of particular value are tables, figures, equations, and definition-type items. Let me know when you think this is ready.

jdebacker:
@rickecon Latest commit adds back the `OGUSAdoc.pdf`.

PR #649, opened 2020-09-24
Fixing equilibrium section

chusloj:
This PR fixes `equilibrium.md` that couldn't be fixed in #644.

@rickecon , there are 2 main problems that prevented us from fixing this section:

1. Because JB is so sensitive to spaces and tabs in the `.md` files, all of the steps for both of the algorithms listed in `equilibrium.md` had to be reformatted so that all text is indented starting when it's flush with the editor. In the current implementation, the body of each section starts 2-3 spaces away from the editor wall (which I understand is intentional). 

2. For the in-line math, when we wrote the following:

```markdown
{some text}

$$
    \math
$$

{some text}
```

We **should have** written:

``` markdown
{some text}

    $$
        \math
    $$

{some text}
```

We just had to indent the each entire in-line math segment one additional time. Also, JB is sensitive to the 2 spaced between {some text} and the indented math block both above and below.

### DIFF
--- a/docs/book/content/theory/equilibrium.md
+++ b/docs/book/content/theory/equilibrium.md
@@ -31,81 +31,92 @@ A non-autarkic stationary steady-state equilibrium in the `OG-USA` model is defi
 (SecEqlbSSsoln)=
 ### Stationary Steady-state Solution Method
 
-  This section describes the solution method for the stationary steady-state equilibrium described in Definition {ref}`DefSSEql`. The steady-state is characterized by $2JS$ equations and $2JS$ unknowns. However, because some of the other equations cannot be solved for analytically and substituted into the Euler equations, we use a fixed point algorithm to solve for the steady-state. We begin by making a guess at steady-state interest rate $\bar{r}$, total bequests $\overline{BQ}$, total household transfers $\overline{TR}$, and income multiplier $factor$. We call these four steady-state variables the "outer loop" variables in our steady-state solution method and the determination of a fixed point over these variables the ``outer loop'' of the steady-state solution method. The outer loop variables are the macroeconomic variables necessary to solve the household's problem.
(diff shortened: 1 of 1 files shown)


chusloj:
**update**: it broke again.

rickecon:
The document to compare against--that we want the OG-USA Jupyter Book documentation to look like--is [`OGUSAdoc.pdf`](docs/OGUSAdoc.pdf). For the `equilibrium.md` chapter, we want it to look like Chapter 14 of [`OGUSAdoc.pdf`](docs/OGUSAdoc.pdf). The goal is to replace that PDF document with the Jupyter Book documentation. The following are obstacles that I still can't make sufficiently good with Jupyter Book.
* Chapter numbering, table numbering, and any object numbering resets within each section, which means that numbered chapters, tables, etc. would have non-unique chapter numbers. An exception to this is figures.
* Tables are really clunky and limited with respect to LaTeX tables. And LaTeX tables aren't even that great. We need something better for tables from Jupyter Book.
* LaTeX has a functionality for repeating equations and keeping the initial number by using the `\tag{\ref{}}` construct. Jupyter Book has no way to do that that I can find.
* We can't get the editable code to produce the image as is shown in the Jupyter Book documentation. This would be nice so that the editable Jupyter notebook version of each page could have editable executable code for producing and changing the figure.
* And the biggest issue with the `equilibrium.md` chapter is nested lists. We can't seem to get a nested list that does the following.

1. Some stuff
  i. some sub stuff related to some stuff
  ii. some more sub stuff related to to some stuff
2. Other stuff
  i. some sub stuff related to other stuff
  ii. some more sub stuff related to to other stuff

cc: @chusloj @jdebacker

jdebacker:
> We can't get the editable code to produce the image as is shown in the Jupyter Book documentation. This would be nice so that the editable Jupyter notebook version of each page could have editable executable code for producing and changing the figure.

These build fine for me locally.  Do they for you two?  I think, then, the issue is something when we build and deploy to GH-pages via the GH action.  

All other issues seem to be current limitations to Jupyter-Book.  e.g., all their examples of tables have quite simple tables with poor aesthetics and all examples of enumerated lists only have Arabic numbers at all levels.  

But with the exception of the chapter/section/equation numbering issues, I can live with the rest.  I think the other advantage of JB dominate the better look of the pdf and I've be ok with replace that pdf if this numbering issue can be sorted.

jdebacker:
Given that the issues outlined above are either limitations of Jupyter-Book or need further exploration (why the JB built via the GH Action doesn't match that build locally - e.g., because of package differences), shall we take the WIP label off this and merge? 

cc @rickecon @chusloj

PR #652, opened 2020-09-27
Create calibration of income process for top percentile groups

jdebacker:
This PR updates `income.py` to create earnings ability profiles that are calibrated to the top 0.1% (excluding the top 0.5%), the top 0.5% (excluding the top 0.1%), the top 0.1% (excluding the top 0.01%) and the top 0.01%.

Without the full population files of administrative data (e.g., from SSA or IRS), one cannot find a panel data with representation of such top earners.  Instead, we make the assumption that the top 3 income groups share the same shape to their lifetime income profiles at the top 1%.  We then use the 2018 updates to [Piketty and Saez (*QJE*, 2003)](https://eml.berkeley.edu/~saez/pikettyqje.pdf) to scale the profiles of the top 1% up to represent these higher percentile groups.

Adding incomes at the very top will help OG-USA better estimate revenue and distributional impacts from policies targeting these very top earners such as those proposed by several 2020 presidential candidates.

PR #662, opened 2020-10-11
SS starting values

jdebacker:
This PR creates kwargs in the `SS.run_SS()` function for the scalar starting values used in the SS solution (`r`, `TR`, `factor`).  This helps with changing starting values when one wants to solve for the SS.

Note that starting values for array variables (`BQ`, `n`, `b`) are not included in the kwargs because these are more complicated to specify.  Also, in my experience, adjusting one of these scalars (usually `r`) does the trick (see comment in PR #652).


jdebacker:
@rickecon I think this is ready for review.  When I look at the code coverage changes [here](https://codecov.io/gh/PSLmodels/OG-USA/compare/44a24d434da92ed2a46cf6d6fa5a35af71a6f48c...e0c3420cccd08d0cc997dcd222bcdd81d3f5e3ab/tree/ogusa), I don't see declines in any files.  The issue with codecov/patch maybe due to the fact that total coverage of `SS.py` on Travis CI is not great (given that we have to leave out some case due to time limits).  But coverage does not go down with the PR.

rickecon:
This looks great. Thanks, @jdebacker.

rickecon:
@jdebacker . If you just resolve the one conflicting block in `./ogusa/tests/test_SS.py`, then this is ready to go.

PR #663, opened 2020-10-14
Add a macro parameters calibration chapter to JB

jdebacker:
This PR adds a skeleton chapter to the Jupyter Book doc that begins to outline the calibration of parameters using macroeconomic data.

PR #665, opened 2020-10-22
Fix test in test_get_micro_data.py

rickecon:
I ran the full set of tests on the OG-USA master branch with most recent commit [d3c38c7f](https://github.com/PSLmodels/OG-USA/commit/d3c38c7fcee65d5942420d8baeb7f86326a80775) (up to PR #663). One test failed in `test_get_micro_data.py` in the [`test_get_calculator_puf()`](https://github.com/PSLmodels/OG-USA/blob/master/ogusa/tests/test_get_micro_data.py#L132) function. The function could not find the `puf.csv` file. The fix was simply to add `PUF_PATH` to the function call. Now it passes.
```
...

ogusa/tests/test_get_micro_data.py ........F.....                        [ 20%]

...

self = <pkg_resources.PathMetadata object at 0x7ff4c29c9430>
manager = <pkg_resources.ResourceManager object at 0x7ff4c28fdac0>
resource_name = 'taxcalc/puf.csv'

    def get_resource_stream(self, manager, resource_name):
>       return open(self._fn(self.module_path, resource_name), 'rb')
E       FileNotFoundError: [Errno 2] No such file or directory: '/Users/richardevans/opt/anaconda3/envs/ogusa-dev/lib/python3.8/site-packages/taxcalc/puf.csv'

../../../../opt/anaconda3/envs/ogusa-dev/lib/python3.8/site-packages/pkg_resources/__init__.py:1623: FileNotFoundError

During handling of the above exception, another exception occurred:

    @pytest.mark.full_run
    def test_get_calculator_puf():
        iit_reform = {
            'II_rt1': {2017: 0.09},
            'II_rt2': {2017: 0.135},
            'II_rt3': {2017: 0.225},
            'II_rt4': {2017: 0.252},
            'II_rt5': {2017: 0.297},
            'II_rt6': {2017: 0.315},
            'II_rt7': {2017: 0.3564}
            }
>       calc = get_micro_data.get_calculator(
            baseline=False, calculator_start_year=2017, reform=iit_reform,
            data=None,
            records_start_year=PUF_START_YEAR)

ogusa/tests/test_get_micro_data.py:142: 
```

cc: @jdebacker

### DIFF
--- a/ogusa/tests/test_get_micro_data.py
+++ b/ogusa/tests/test_get_micro_data.py
@@ -141,8 +141,7 @@ def test_get_calculator_puf():
         }
     calc = get_micro_data.get_calculator(
         baseline=False, calculator_start_year=2017, reform=iit_reform,
-        data=None,
-        records_start_year=PUF_START_YEAR)
+        data=PUF_PATH, records_start_year=PUF_START_YEAR)
     assert calc.current_year == 2013
 
 


rickecon:
@jdebacker . I have placed this as a `[WIP]` work in progress because I also have all the tests in `/regression/test_regression.py` either failing (56) or erroring because of FileNotFound (288). I don't have the `./regression/regression_results/` folder with all of the requisite content.

jdebacker:
@rickecon -- this is a good catch.  I was not seeing a failure of this because OG-USA had been using the Tax-Calculator `Records` class to locate the PUF in the case of `data=None` (see [here](https://github.com/PSLmodels/OG-USA/blob/90b23d3aa98e12dde1a1e5118877909b963666b0/ogusa/get_micro_data.py#L56) and [here](https://github.com/PSLmodels/Tax-Calculator/blob/32c0f6d5e0642ae27ce7d891fa5dcf54c5887085/taxcalc/records.py#L120), but I guess that only works if someone has the PUF in a place the `Records` class is looking.

Re the regression tests: These will all fail when current results are compared against the cached results.  We *could* commit those cached results to the repo and then update them with the new model runs.  On the other hand, this maybe a good time to change how those tests are setup.  Currently, the baseline policy + 10 reform policies are run and all results (SS, TPI, model parameters) are compared to the cached results (which are stored in 33 different pickle files).  This is a lot of additional binary files to have in the repo.  We should probably open an issue to discuss how we can consolidate these tests/results and in what format we should save them (e.g., csv might be preferred for security reasons and to more easily track changes over time).

rickecon:
@jdebacker . OK. Then this PR is ready to go, and I will open the issue you mentioned above.

Issue #673, opened 2020-10-29
Tests failing with ParamTools 0.16.0


jdebacker:
Tests are failing with ParamTools 0.16.0.  Model seems to run fine, but there is an incompatibility with cached `Specifications` objects used in tests (created with an older version of ParamTools) and the new version.

We'll need to figure out how to have old object conform with new ParamTools, or save new `Specifications` objects for testing.

hdoupe:
@jdebacker Sorry about this. I tested against Tax-Calculator, CCC, Tax-Brain, and Tax-Cruncher. For the sake of my computer, I skipped OG-USA. What are the error messages?

jdebacker:
@hdoupe I wouldn't expect you to test against OG-USA!  This will happen, but I appreciate you chiming in here as maybe you can suggest a quick fix.

Here's the traceback:
```
____________ ERROR collecting ogusa/tests/test_parameter_tables.py _____________
66
ImportError while importing test module '/home/runner/work/OG-USA/OG-USA/ogusa/tests/test_parameter_tables.py'.
67
Hint: make sure your test modules/packages have valid Python names.
68
Traceback:
69
/usr/share/miniconda/envs/ogusa-dev/lib/python3.7/importlib/__init__.py:127: in import_module
70
    return _bootstrap._gcd_import(name[level:], package, level)
71
ogusa/tests/test_parameter_tables.py:15: in <module>
72
    os.path.join(CUR_PATH, 'test_io_data', 'model_params_baseline.pkl'))
73
ogusa/utils.py:364: in safe_read_pickle
74
    obj = pickle.load(f, encoding='latin1')
75
E   ModuleNotFoundError: No module named 'paramtools.tree'
```

hdoupe:
Oh, the `tree` module was removed in 0.16.0...you may be able to dump the parameter values and load them with the new version of paramtools:

```
conda install "paramtools==0.15.1" -c conda-forge
```

```python
obj = pickle.load(f, encoding='latin1')

with open("model_params_baseline.json", "w") as f:
    f.write(
        json.dumps(obj.specification(use_state=False, meta_data=False, serializable=True))
    )

```

and then:

```
conda install "paramtools==0.16.0" -c conda-forge
```

```python

params = Specifications() 

params.adjust("model_params_baseline.json")

```

However, this will only work if all parameter values are updated using `adjust`. I think there are some places where OG-USA just updates the parameter value directly like this:

[...]

PR #674, opened 2020-10-30
Update pickled Specifications objects

jdebacker:
This PR updates the cached `Specifications` objects used in testing, which are causing the failures noted in Issue #673.


jdebacker:
Added in progress label.  Will need to update the inputs and outputs for `test_TPI.test_twist_doughnut`.

jdebacker:
Ok, I believe all files with cached `Specifications` objects have been updated.   I had to update the results for `test_twist_doughnut` to be consistent with the Tax-Calculator 3.0 and the latest OG-USA.

jdebacker:
Tests are passing on Travis CI, but not on the GitHub Action, which runs the same set of tests.  Error there is due to issue reading in one of the new pickle files.  See [here](https://github.com/PSLmodels/OG-USA/pull/674/checks?check_run_id=1336373890#step:5:55):

```
____________ ERROR collecting ogusa/tests/test_parameter_tables.py _____________
51
ogusa/tests/test_parameter_tables.py:15: in <module>
52
    os.path.join(CUR_PATH, 'test_io_data', 'model_params_baseline.pkl'))
53
ogusa/utils.py:364: in safe_read_pickle
54
    obj = pickle.load(f, encoding='latin1')
55
E   ValueError: unsupported pickle protocol: 5
```

Note that the pickle was saved with `cloudpickle`.  It's rare, but I've run into a similar error before, usually with pickle files shared on Google Drive, so I thought it was an issue with Drive's version control interacting with the pickle files.  I never found a fix - I just had to create new files and eventually the error would go away.  

cc @hdoupe

jdebacker:
The decline in code coverage is on the `test_TPI.py` script.  Again, I think these test files should be excluded from the coverage reports, but I don't know what to do besides what we have in [`codecov.yml`](https://github.com/PSLmodels/OG-USA/blob/master/codecov.yml).

jdebacker:
Ok, this may be an issue with saving a pickle in Python 3.8 and trying to read it in Python 3.7.  I tried saving the pickle files (in Python 3.8) with `protocol=4`, which I thought would be compatible with Python 3.7.  But an error is still returned (albeit different):

```
_pickle.UnpicklingError: could not find MARK
```

Any other ways around this?  It just seems to matter for the `ParamTools.parameters` object saved with `cloudpickle`.

cc @hdoupe @rickecon

Issue #623, opened 2020-08-18
Large file storage


jdebacker:
To run the full suite of tests for OG-USA one needs two files that cannot be checked into the repo:
1. `puf.csv` -- cannot be checked in due to confidentiality of the data.
2. `ogusa/tests/test_io_data/tax_func_loop_inputs_large.pkl` -- cannot be checked in because the file size (~124MB) exceeds the 100MB limit on GitHub.

If we want to version control (2), we could look into [Git Large File Storage Options](https://docs.github.com/en/github/managing-large-files/about-git-large-file-storage).

If we just want others to be able to access this file, we could have the test calling this data file reach out to the file stored on the web.  Some options for storage:
1. Dropbox
2. Google Drive
3. Jason's personal website
4. An Amazon s3 bucket

The additional cost of the 1st 3 is zero, but performance (e.g., download speeds) maybe best with (4)?

cc @rickecon

rickecon:
@jdebacker thanks for this clarifying issue. I think we should have three tiers of unit tests:

1. `CI` tests: tests that run upon commit to master of main repository
2. `full_run_a` tests: Includes all the CI tests plus all the non-CI tests minus any tests that use `puf.csv`.
3. `full_run_b` tests: Includes all the tests: `full_run_a` + tests that use `puf.csv`.

For now, I agree with your comment on the PSL Community call today, that it is nice to have all the testing data files as part of the repository. And it would be nice if we could get away with including those testing datafiles in the forthcoming conda package.

For `full_run_a` tests, we could post the `tax_func_loop_inputs_large.pkl` file in a Google Drive folder and have the tests that require that file download it using the `requests` library. This approach is clunkier than the Amazon s3 bucket, but it is only one file and should be fine. It would also be nice to specify the tests that use that file to search for it in an ignored folder on the local hard drive first, such as `ogusa/tests/test_io_data/ignored/tax_func_loop_inputs_large.pkl`. That way, we could include that folder in the `.gitignore`, but we could also run the `full_run_a` tests without having to wait for a  125MB download.

Finally, the `full_run_b` tests will search for the `puf.csv` on the local hard drive in the default `ogusa/puf.csv` path. The alternative would be to have all the testing files and the model on a secure server, where the user could run the test on the confidential data and only see the output. But I don't think we want to do that right now.

jdebacker:
@rickecon thanks for the reply.  I like your suggestions, but a few idea/clarifications:

[...]

PR #688, opened 2020-12-11
Compress large file

jdebacker:
This PR addressed Issue #623.  It does so by compressing the pickle file used in the `test_txfunc.py::test_tax_func_loop` unit test using the `bz2` package.  

Thanks to @chusloj for this suggestion.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -28,5 +28,5 @@ dependencies:
 - pip:
   - jupyter-book>=0.8.0
   - cs-kit
-  - xlrd
+  - xlrd==1.2.0
   - cs2tc
--- a/ogusa/tests/test_txfunc.py
+++ b/ogusa/tests/test_txfunc.py
@@ -184,6 +184,9 @@ def test_tax_data_sample():
     assert isinstance(df, pd.DataFrame)
 
 
+@pytest.mark.full_run
+# mark as full run since results work on Mac, but differ on other
+# platforms
 def test_tax_func_loop():
     '''
     Test txfunc.tax_func_loop() function.  The test is that given
--- a/ogusa/tests/test_txfunc.py
+++ b/ogusa/tests/test_txfunc.py
@@ -1,10 +1,11 @@
 from ogusa import txfunc
-import multiprocessing
 from distributed import Client, LocalCluster
 import pytest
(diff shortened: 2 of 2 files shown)


rickecon:
@jdebacker. This looks good to me. Is it ready to go from your end?

jdebacker:
@rickecon Yes - the test affected still takes to long for Travis CI, so it's still flagged not to run here.  But it did pass locally.

But I now see that Travis CI has disappeared here.  Shall I include this test in our GH testing suite and see if the GH Action has a timeout limit?

rickecon:
@jdebacker . Yes. I think we should try to add this test to the GH Action testing and see if it goes through. Testing the tax functions seems like an important check for every PR.

Regarding TravisCI, are you saying that it just stopped working? That is weird. But I guess we're OK to just do our testing through GH Actions.

chusloj:
> see if the GH Action has a timeout limit

@jdebacker The GH timeout limit for a job within a workflow is [6 hours](https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration#usage-limits).

jdebacker:
>  I found that the GH timeout limit for a job within a workflow is 6 hours.

Wow!  That is great.  Thanks for letting us know @chusloj.

jdebacker:
@rickecon There was an issue with running the `test_tax_func_loop` on GH Actions.  The results `txfunc.txfunc_est`, or any function using it, are dependent on the platform this optimization routine is run on due to differences in the `MKL` package.  So I've added back the `full_run` flag for this test. 

But we should start to add tests to the suite of tests run in our GH Actions now that we are not using Travis CI.

FYI, it [looks like Travis CI may still be free](https://github.com/marketplace/travis-ci) for us, but they must have changed something with their API because I noticed that it stopped running on the Cost-of-Captial-Calculator repo last week and OG-USA this week.

jdebacker:
Looks like the `xlrd` packages was updated for the first time in a couple years and it broke our tests that read Excel files from the web.  To solve, I followed the solution offered [here](https://github.com/equinor/fmu-tools/issues/88) and just pinned the version of the `xlrd` package.

PR #693, opened 2020-12-29
Function to compute percentiles

jdebacker:
This PR adds a function to the `Inequality` class to compute the value of a variable at a given percentile in the distribution.


jdebacker:

PR #695, opened 2021-01-07
Fix indexing issue with linear tax rate function

jdebacker:
This PR fixes an issue with the indexing of parameters for the linear tax rate functions.  An index was not properly updated when changes were made to the tax rate functions in PR #629.

Tests are also added to cover all cases in the `txfunc.txfunc_est()` function.


jdebacker:

rickecon:
@jdebacker. Ah, Jason, this is the reason why my current runs of the TPC model with `J=10` are not working. I deserve that for being slow on the review for this.

PR #697, opened 2021-01-11
Fix user input of initial_debt_ratio on C/S

jdebacker:
This PR updates `functions.py` so that the initial debt to GDP ratio input by the user will be applied to the baseline run of the model. 

NB that this parameter does not affect the reform run, which takes the level of debt from the baseline ([see lines 70-73 of `fiscal.py`](https://github.com/PSLmodels/OG-USA/blob/45b8c2d31f8d88d68a0fad8b281077321c386d97/ogusa/fiscal.py#L70)).

### DIFF
--- a/cs-config/cs_config/functions.py
+++ b/cs-config/cs_config/functions.py
@@ -156,8 +156,8 @@ def run_model(meta_param_dict, adjustment):
         utils.mkdirs(_dir)
 
     # Dask parmeters
-    client = Client()
-    num_workers = 5
+    client = None#Client()
+    num_workers = 1#5
     # TODO: Swap to these parameters when able to specify tax function
     # and model workers separately
     # num_workers_txf = 5
@@ -173,8 +173,9 @@ def run_model(meta_param_dict, adjustment):
     constant_param_set = {
         'frisch', 'beta_annual', 'sigma', 'g_y_annual', 'gamma',
         'epsilon', 'Z', 'delta_annual', 'small_open', 'world_int_rate',
-        'initial_foreign_debt_ratio', 'zeta_D', 'zeta_K', 'tG1', 'tG2',
-        'rho_G', 'debt_ratio_ss', 'budget_balance'}
+        'initial_debt_ratio', 'initial_foreign_debt_ratio', 'zeta_D',
+        'zeta_K', 'tG1', 'tG2', 'rho_G', 'debt_ratio_ss',
+        'budget_balance'}
     filtered_ogusa_params = OrderedDict()
     for k, v in adjustment['OG-USA Parameters'].items():
         if k in constant_param_set:
@@ -187,7 +188,7 @@ def run_model(meta_param_dict, adjustment):
         OGDIR = os.path.dirname(OGPATH)
         tax_func_path = None#os.path.join(OGDIR, 'data', 'tax_functions',
                         #             cached_pickle)
-        run_micro_baseline = False
+        run_micro_baseline = True
     else:
         tax_func_path = None
         run_micro_baseline = True
@@ -241,7 +242,7 @@ def run_model(meta_param_dict, adjustment):
 
     # Shut down client and make sure all of its references are
     # cleaned up.
-    client.close()
+    # client.close()
     del client
 
     return comp_dict
--- a/cs-config/cs_config/functions.py
+++ b/cs-config/cs_config/functions.py
@@ -156,8 +156,8 @@ def run_model(meta_param_dict, adjustment):
         utils.mkdirs(_dir)
 
     # Dask parmeters
-    client = None#Client()
-    num_workers = 1#5
+    client = Client()
+    num_workers = 5
     # TODO: Swap to these parameters when able to specify tax function
     # and model workers separately
     # num_workers_txf = 5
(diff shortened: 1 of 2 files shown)


jdebacker:
Adding WIP.  Change didn't have anticipated effect.

jdebacker:
Have now tested that changes work as expected (see updates to the `write-output.py` script in this PR).

@rickecon  This PR is ready for your review.

cs-config/cs_config/functions.py
```diff
@@ -187,7 +188,7 @@ def run_model(meta_param_dict, adjustment):
         OGDIR = os.path.dirname(OGPATH)
         tax_func_path = None#os.path.join(OGDIR, 'data', 'tax_functions',
                         #             cached_pickle)
-        run_micro_baseline = False
+        run_micro_baseline = True
```
rickecon, line 191: @jdebacker . It doesn't really matter, but it seems like for cleanliness and consistency that we should move this statement `run_micro_baseline = True` outside of the `if`/`else` statement given that it is `True` in both cases now. Also, I can't figure out what this change does substantially.

jdebacker: @rickecon The whole if/else here is not operational since we haven't used the cached pickle file.  By passing `tax_func_path=None`, the model would estimate the tax functions regardless of the boolean being `True` or `False`.  Thus this change has no substantive effect.

I thought we might as well leave in the if/else for now, even if not operational, because I do think we should get something like this working (see Issue #698).  Right now our cached tax functions are DEP, age-specific functions and the C/S app only runs with DEP, non-age-specific, which is why the cached functions aren't used here.

rickecon:
@jdebacker. The problem was that the `initial_debt_ratio` variable changed the reform and not the baseline in C/S, but the code sets the reform value of `initial_debt_ratio` to whatever the baseline was. So the user input made a change that was both the wrong change (change the reform and not the baseline) and could not be implemented (reform level was required to equal the baseline). It is not clear to me from the changelog what this PR does.

One implementation specific to Compute.Studio would be to have the input of that particular `initial_debt_ratio` field change the value in both the baseline and reform. However, more generally inside of more sophisticated OG-USA runs on one's local machine, I could imagine a reform in which the `initial_debt_ratio` value were different in the baseline and in the reform.

We need to implement at least the C/S functionality from the first paragraph. But it would also be nice to implement the richer functionality described in the second paragraph.

cs-config/cs_config/functions.py
rickecon: @jdebacker OK.

cs-config/cs_config/functions.py
```diff
@@ -187,7 +188,7 @@ def run_model(meta_param_dict, adjustment):
         OGDIR = os.path.dirname(OGPATH)
         tax_func_path = None#os.path.join(OGDIR, 'data', 'tax_functions',
```
rickecon, line 189: @jdebacker Does this create an error on line 189 to not have a space (prefer two spaces) between `None` and the commented out code `#os.path.join(...`?

jdebacker: No error in execution.  This would not pass a `pycodestyle` check, though.

jdebacker:
> @jdebacker. The problem was that the initial_debt_ratio variable changed the reform and not the baseline in C/S, but the code sets the reform value of initial_debt_ratio to whatever the baseline was. So the user input made a change that was both the wrong change (change the reform and not the baseline) and could not be implemented (reform level was required to equal the baseline). It is not clear to me from the changelog what this PR does.

Changes to the parameters input in the C/S app apply to the *reform*, unless specified otherwise.  The list of parameters in the `constant_param_set` set specified on line [173 of `functions.py`](https://github.com/PSLmodels/OG-USA/blob/45b8c2d31f8d88d68a0fad8b281077321c386d97/cs-config/cs_config/functions.py#L173) makes sure that values of those parameters, if changed in the C/S app, apply to both the baseline and reform run.  This  PR adds `initial_debt_ratio` to that list. 

Note that the `initial_debt_ratio` is a parameter that *only* affects the baseline.  In the reform simulation, the initial amount of debt (not the ratio to GDP) is set to the initial amount found in the baseline run (see [lines 70-73 of `fiscal.py`](https://github.com/PSLmodels/OG-USA/blob/45b8c2d31f8d88d68a0fad8b281077321c386d97/ogusa/fiscal.py#L70)).  Thus, this PR has the intended effects - if the user changes the initial debt-to-GDP level in the app, it applies that to the baseline run (it does also apply that to the reform run, but it has no effect there since in the reform, the level of D is exogenous, but the ratio D/Y is endogenous).

> One implementation specific to Compute.Studio would be to have the input of that particular initial_debt_ratio field change the value in both the baseline and reform. However, more generally inside of more sophisticated OG-USA runs on one's local machine, I could imagine a reform in which the initial_debt_ratio value were different in the baseline and in the reform.

As noted above, OG-USA does not include the ability to specify different initial debt-to-GDP ratios for the reform and baseline simulations.  We could make the model more flexible in this regard.  But right now the model is built off the premise that the initial conditions  (distribution of wealth, government debt) are set at beginning of the time path and that we take these as the starting point and then simulate what happens from this point onwards under alternative policy assumptions. 

> We need to implement at least the C/S functionality from the first paragraph. But it would also be nice to implement the richer functionality described in the second paragraph.

This PR achieves the first goal.  Richer functionality is left to a future PR.

PR #705, opened 2021-01-30
Add test and build for python 3.9

jdebacker:
This PR tests OG-USA for Python 3.9.


rickecon:
@jdebacker. This is great. I think we should get rid of testing for Python 3.7. I think a good rule is to only have testing for two versions of Python at a time.

PR #736, opened 2021-07-22
Fix typo in docs

MaxGhenis:
One `zeta` missing a `\`


rickecon:
@MaxGhenis . Never has the "\" meant so much to so many. Thx.

PR #745, opened 2021-08-25
Update OG-Core documentation

rickecon:
This PR updates the documentation of OG-Core.
* Adds core maintainers section to `README.md`
* Updates the version in `intro.md`
* Removes references to Tax-Calculator model in OG-Core documentation
* Fixes broken equations in text and removes broken equations that refer to other documentation
* Add `demographics.md` section without calibration discussion.
* Updated the references and citations location of `.bib` files and `.md` files

@jdebacker


docs/book/content/OGCore_references.md
```diff
@@ -0,0 +1,5 @@
+# References
+
+```{bibliography} ../OGCore_references.bib
+:style: alpha
```
jdebacker, line 4: @rickecon What do you think about the `plain` style?  It seems to avoid warnings when a paper has the same authors and year (which results in a duplicate alpha-numeric label).

rickecon: @jdebacker. LaTeX doesn't choke as long as you give the paper a unique reference ID. I wish the Jupyter Book processing did that as well. But I am OK with this citation approach. However, I think the best option is to have the last names of the authors spelled out in the reference, and I don't think this is an option in Jupyter Book.

jdebacker:
Thanks @rickecon!

PR #750, opened 2021-09-22
Financial Intermediary

jdebacker:
This PR changes how one thinks about how domestic and foreign savings are held.  It introduces a financial intermediary into the model. This intermediary then rationalizes the return to the portfolio of government debt held by domestic and foreign investors.  

The changes in this PR do not alter any of the equations underlying the model or its solution, only their interpretation.

All references to `r_hh` in the theory and code have been changed to `r_p` to better denote the return to the portfolio of investments held by the financial intermediary.  This return is realized not just by domestic households, but also foreign investors in domestic assets.

### DIFF
--- a/.github/workflows/deploy_docs.yml
+++ b/.github/workflows/deploy_docs.yml
@@ -24,8 +24,6 @@ jobs:
       - name: Build # Build Jupyter Book
         shell: bash -l {0}
         run: |
-          pip install jupyter-book
-          pip install sphinxcontrib-bibtex>=2.0.0
           pip install -e .
           cd docs
           jb build ./book
--- a/.github/workflows/deploy_docs.yml
+++ b/.github/workflows/deploy_docs.yml
@@ -25,7 +25,7 @@ jobs:
         shell: bash -l {0}
         run: |
           pip install jupyter-book
-          pip install sphinxcontrib-bibtex==1.0.0
+          pip install sphinxcontrib-bibtex>=2.0.0
           pip install -e .
           cd docs
           jb build ./book
--- a/.github/workflows/docs_check.yml
+++ b/.github/workflows/docs_check.yml
@@ -22,7 +22,7 @@ jobs:
         shell: bash -l {0}
         run: |
           pip install jupyter-book
-          pip install sphinxcontrib-bibtex==1.0.0
+          pip install sphinxcontrib-bibtex>=2.0.0
           pip install -e .
           cd docs
           jb build ./book
--- a/.github/workflows/docs_check.yml
+++ b/.github/workflows/docs_check.yml
@@ -21,8 +21,6 @@ jobs:
       - name: Build # Build Jupyter Book
         shell: bash -l {0}
         run: |
-          pip install jupyter-book
-          pip install sphinxcontrib-bibtex>=2.0.0
           pip install -e .
           cd docs
           jb build ./book
--- a/docs/book/OGCore_references.bib
+++ b/docs/book/OGCore_references.bib
@@ -133,6 +133,16 @@
   edition = {4th edition},
 }
 
+@TECHREPORT{BenzellEtAl:2017,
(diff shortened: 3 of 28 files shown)


rickecon:
@jdebacker. This PR looks great. I have reviewed 12 of the 20 changed files. I am currently going through the documentation changes, comparing this updated documentation to the current documentation. While I am finishing the documentation review, maybe you can answer the following question.
* In `text_SS.py`, `test_TPI.py`, `test_output_plots.py`, and `test_output_tables.py` you create a new dictionary item `r_p` from the `r_hh` dictionary item without removing the `r_hh` dictionary item. Shouldn't we just rename all the `r_hh` dictionary items in the testing `pkl` files so we are not carrying around vestigial `r_hh` items in these pickle files?

jdebacker:
@rickecon asks:

> In text_SS.py, test_TPI.py, test_output_plots.py, and test_output_tables.py you create a new dictionary item r_p from the r_hh dictionary item without removing the r_hh dictionary item. Shouldn't we just rename all the r_hh dictionary items in the testing pkl files so we are not carrying around vestigial r_hh items in these pickle files?

Yes, I wanted just rename the dictionary keys so that I could show that no test results had to change.  This does make the tests a little less clean than would be the case if I just replaced the cached `pkl` files.  But I think it's helpful to show that no results needed to change for tests to pass. 

My recommendation for stuff like this is to not replace the cached files (if possible) on PRs of substance.  Then, periodically and in separate PRs that just update the cached files, update those cached files and clean up the `test_*.py` modules as necessary to remove things like the renaming of dictionary keys done here.

Glad to hear suggestions on this, though.

rickecon:
@jdebacker. I think it is going to be hard to keep track of `.pkl` file cleanup that needs to happen across multiple PRs. The way you coded those four files in this PR already shows that none of the pickle files had to change. My recommendation now is to update the pickle files so we don't have to keep track of the need to remove the `r_hh` item in some future cleanup.

I agree that there are at least two ways to do this--your proposed method of doing big `.pkl` cleanups after major changes, and my proposed method of updating the `.pkl`s with each respective PR. My preference is the latter, but I am happy to go with this if you feel anything more than indifferent between the two.

jdebacker:
> I think it is going to be hard to keep track of .pkl file cleanup that needs to happen across multiple PRs.

I don't think it will be.  When one sees the `test_*.py` models having many extra lines to rename or exclude variables, etc. then it'll be obvious that it's time for some housekeeping.

A difficult thing would be if I committed the changed pickle files to the PR before you confirmed tests pass with the unchanged files.  It'd be hard to go backwards on that if substantive work were done on subsequent commits.

rickecon:
OK. I'll check those `test_*.py` files off in my review. @jdebacker

rickecon:
@jdebacker. In reading the documentation, I have noticed that none of the BibTeX references are compiling, and this is true in the current documentation (see the first paragraph of the [Demographics](https://pslmodels.github.io/OG-Core/content/theory/demographics.html) chapter, any other BibTeX reference in the text and the [References](https://pslmodels.github.io/OG-Core/content/OGCore_references.html) chapter). I am currently working on fixing this as well.

rickecon:
@jdebacker. I deleted my last comment. The BibTeX references ARE compiling, linking, and listing when I compile the Jupyter Book documentation in this PR locally on my machine. However, they are not compiling in the current documentation online (see the first paragraph of the [Demographics](https://pslmodels.github.io/OG-Core/content/theory/demographics.html) chapter, any other BibTeX reference in the text, or the empty [References](https://pslmodels.github.io/OG-Core/content/OGCore_references.html) chapter). I don't know why this is happening.

rickecon:
@jdebacker. I submitted [a PR](https://github.com/jdebacker/OG-Core/pull/22) to your branch that updates some things I found in my review. Once this is merged in and passes all tests, I am ready to merge this PR.

rickecon:
@jdebacker. Thanks for this PR. LGTM.

Issue #53, opened 2015-06-16
Add flags in demographics, income, labor, and wealth_data.py for graphs


evan-magnusson:
We don't want the graphs generated in regular runs, so add flags to each graphing function in demographics and income to prevent this from happening.  Also, in labor and wealth, make functions that control graphing.  In labor, write a function that produces the data array.

evan-magnusson:
Finished.  Added a flag variable in run_simulation.py to control all graphing functions in these files.

PR #762, opened 2021-12-02
Added taxcalc_version to txfunc.py pickle

rickecon:
This PR adds `taxcalc_version` to the pickle file output from the `tax_func_estimate()` function at the end of the `txfunc.py` module. This is helpful for the OG-USA calibration case in which estimated tax function parameters are taken from a previously estimated and saved file. See [Issue #53](https://github.com/PSLmodels/OG-USA/issues/53) and [PR #54](https://github.com/PSLmodels/OG-USA/pull/54) in the OG-USA repository.

cc: @jdebacker

### DIFF
--- a/ogcore/txfunc.py
+++ b/ogcore/txfunc.py
@@ -16,6 +16,7 @@ import scipy.optimize as opt
 from dask import delayed, compute
 import dask.multiprocessing
 import pickle
+import pkg_resources
 import ogcore.parameter_plots as pp
 from ogcore.constants import DEFAULT_START_YEAR, SHOW_RUNTIME
 import warnings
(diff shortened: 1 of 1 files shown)


jdebacker:
@rickecon `taxcalc` is not a dependency of OG-Core.  I believe the retrieval of the microsimulation model version should be something for the country-specific packages since they will each depend on a different microsimulation model.

rickecon:
@jdebacker. Ah, yes. You're right. We don't want to do this in OG-Core. This should be done in the country-specific repos. But this is a harder problem than you think. The pickle files of estimated tax functions `TxFuncEst_baseline.pkl` and `TxFuncEst_policy.pkl` are saved to hard drive by the `txfunc.py` module in OG-Core. So any saved tax function estimation pickle file will not have the Tax-Calculator version. The version is outputted of course in the `OG-USA` modules `calibrate.py` and `get_micro_data.py` when you run the estimation. However, that version information is not saved to the tax func parameter pickle. I think the answer is to get rid of the version requirement on [lines 316-326](https://github.com/PSLmodels/OG-USA/blob/master/ogusa/calibrate.py#L316) in `calibrate.py` when reading the tax function parameter pickle from a previously saved run.

PR #763, opened 2021-12-03
Adjust max for mtry_params in default_parameters.json

rickecon:
This PR increases the maximum value allowed for `mtry_params` values from 10000 to 20000 in line 3308 of [`default_parameters.json`](https://github.com/PSLmodels/OG-Core/blob/master/ogcore/default_parameters.json).

In doing an experiment with the OG-USA repository, I ran Tax-Calculator with a small increase (4%) in all seven personal income tax marginal tax rates, then estimated DEP tax functions. The maximum value set for all `ETR`, `MTRx`, and `MTRy` is 10000. However, my estimated `mtry_params` tax functions had one parameter value ~16000 on the B coefficient (`mtry_params[17, 5, 1]`, the coefficient on the linear term on mtrx quadratic) and one parameter value  ~234 on the D coefficient (`mtry_params[42, 6, 3]`, the coefficient on the linear term on the mtry quadratic).

In `txfunc.py` we check for outliers in the estimated age-specific DEP tax functions based on sum of squared errors or fit, not based on absolute coefficient size. In the first case listed above (`mtry_params[17, 5, 1] ~ 16000`), the sum of squared errors on the fit of the estimated function was only 0.95 standard deviations above the mean sum of squared errors for all 800 estimated DEP age-specific mtry functions. In the other case (`mtry_params[42, 6, 3] ~ 234`), the sum of squared errors on the fit of the estimated function was only 0.03 standard deviations above the mean sum of squared errors for all 800 estimated DEP age-specific mtry functions.

This tells me that the tax functions being estimated are doing a good job of fitting the data. As such, I don't want to truncate or interpolate parameter values like we do in the `find_outliers` and `replace_outliers` functions in the `txfunc.py` module. Instead, I simply change the maximum parameter value for the `mtry_params` object to 20000.

cc: @jdebacker


jdebacker:
@rickecon The one failing test, `test_txfunc.py::test_txfunc_est_on_GH['GS']`, seems to be a result of changes in more recent versions of the package dependencies.  This test passed for me locally with a stale `ogcore-dev` environment.  When I updated all packages, I get the same failure as on GH Actions.  You might just change the expected results for this test to reflect the new parameter estimates that come out of the optimization using the latest version of SciPy.

After that, let's merge this PR.

rickecon:
@jdebacker. Dang. I literally changed one number in the `default_parameter.json`, and that number is a maximum value constraint that should not bind in any test cases. Alright. I’ll update the test data.

jdebacker:
@rickecon It's not the number that's causing the test failure. It's the scipy version.

jdebacker:
BTW, when I run the full set of tests in `test_txfunc.py` locally (after updating packages in my `ogcore-dev` environment), I get 4 test failures:

```
FAILED tests/test_txfunc.py::test_txfunc_est_on_GH[GS] - assert False
FAILED tests/test_txfunc.py::test_tax_func_estimate - assert False
```

rickecon:
@jdebacker . Ugh. OK.

rickecon:
@jdebacker. All tests pass now due to PR #766. This PR is ready now.

PR #722, opened 2021-03-31
Infrastructure

jdebacker:
This PR begins to add infrastructure spending to OG-USA through the inclusion of a public capital good in firms' production functions.

PR #767, opened 2021-12-16
Making root finders more flexible

jdebacker:
This PR replaces calls to `scipy.optimize.fsolve` with calls to `scipy.optimize.root` throughout the code.  This will allow for an option to specify alternative root finding algorithms that may prove more robust in some cases.  With the default method (`method='hybr'`), `opt.root` is equivalent to `opt.fsolve`.


jdebacker:
@rickecon What do you think about parameters in the `default_parameters.json` file that allow the user to select the root finding algorithm for the SS solution's outer loop and a separate parameter for the root finding method to use when solving the household's problem?

rickecon:
@jdebacker . Yes. I love this idea.

jdebacker:
GH Actions test failures are the same issues noted in PR #722.  Otherwise, this PR is ready.

rickecon:
@jdebacker . I cloned your fork to a branch on my repo and tested this out using modifications of the `run_ogcore_example.py` script. I really only worked with the baseline model to solve in the example. But I did six runs, each of which I stopped short. I first made sure the steady-state solved correctly. Then I made sure the time path iteration ran for a few iterations in which the distance metric was converging.

To run my tests, I updated the `og_spec` declaration in `run_ogcore_example.py` to update different versions of the root finder (see the last line).
```python
og_spec = {
        'frisch': 0.41, 'start_year': START_YEAR, 'cit_rate': [0.21],
        'debt_ratio_ss': 1.0, 'alpha_T': alpha_T.tolist(),
        'alpha_G': alpha_G.tolist(),
        'SS_root_method': 'hybr', 'FOC_root_method': 'hybr'}
```

The following are the seven cases I checked. `Yes` means it worked. `No` means it broke or didn't solve.
```
SS_root_method | FOC_root_method | Yes |  No | Description
---------------+-----------------+-----+-----+------------
     hybr      |      hybr       |  X  |     |
      lm       |      hybr       |  X  |     | significantly slower
   broyden1    |      hybr       |     |  X  | broke in SS: infs and NaNs
   broyden2    |      hybr       |     |  X  | broke in SS: infs and NaNs
     hybr      |       lm        |  X  |     | SS solved. Took forever to solve each iteration of TPI
     hybr      |    broyden1     |     |  X  | broke in SS. No progress in SS for many iterations
     hybr      |    broyden2     |     |  X  | broke in SS. No progress in SS for many iterations
```
I didn't check all the other options (`anderson`, `linearmixing`, `diagbroyden`, `excitingmixing`, `krylov`, `df-sane`), but I suspect that all we need right now is `hybr`. It is nice to switch from `scipy.optimize.fsolve` to `scipy.optimize.root`. Then the option to use these other methods is available. We just need to say somewhere in the documentation or in `default_parameters.json` that all the other methods might require further arguments to be input in the code and maybe adjusted starting values.

An alternative that I don't like is to remove the validator choices from `default_parameters.json` that don't work without more arguments or starting value changes. In the example above, this would mean removing `broyden1` and `broyden2` and any other methods that don't work from the `default_parameters.json` validators. I think I would rather just leave them in and allow the option to further customize. The big benefit for me is moving to `scipy.optimize.root`.

rickecon:
@jdebacker. I have reviewed the code changes in this PR, and I am ready to merge it as soon as I confirm that you agree with my assessment in my last comment in this PR about the methods that don't work.

jdebacker:
@rickecon Awesome - thank you for testing these combinations of root finding methods -- this will be good to reference in the future.  I just updated this PR by adding some notes to these new methods parameters in `default_parameters.json` to help address your points.

Issue #770, opened 2022-01-05
Change and unify where pytest output is saved


rickecon:
Currently, when the full set of tests from the `pytest` package is run, the output is saved in multiple places, some of which are in the main directory. The following is the list of `pytest` saved output from a recent full run of the unit tests.
```
OUTPUT_BASELINE/
ogcore/tests/OUTPUT_REFORM_/
ogcore/tests/TPI/
ogcore/tests/test_out.pkl
output.csv
output.json
output.tex
output.xlsx
```
I propose that all testing output be saved in a folder in the `OG-Core/ogcore/tests/` directory that has a name based off of the timestamp when the tests were begun. For example, the directory for a test run today at 1:50p local computer time might be `2022-01-05-1350`. The status check of the saved output would be the following.
```
ogcore/tests/2022-01-05-1350/OUTPUT_BASELINE/
ogcore/tests/2022-01-05-1350/OUTPUT_REFORM_/
ogcore/tests/2022-01-05-1350/TPI/
ogcore/tests/2022-01-05-1350/test_out.pkl
ogcore/tests/2022-01-05-1350/output.csv
ogcore/tests/2022-01-05-1350/output.json
ogcore/tests/2022-01-05-1350/output.tex
ogcore/tests/2022-01-05-1350/output.xlsx
```
This would make testing run output easy to save, archive, and erase.

cc: @jdebacker

jdebacker:
@rickecon Thanks for pointing this out.  A number of files (and perhaps all) created are of no value afterward.  We could use the [`pytest` `tmpdir`](https://docs.pytest.org/en/6.2.x/tmpdir.html) fixture to save these to a temporary directory.  This is done with some other tests in OG-Core, such as [test_parameter_plots.test_plot_mort_rates_save_fig](https://github.com/PSLmodels/OG-Core/blob/f0e241690ce64616843ab82ebb4e5fbe038ec9ec/ogcore/tests/test_parameter_plots.py#L46

rickecon:
This Issue is resolved by PR #775.

PR #775, opened 2022-01-19
Update tests to save all output to temporary files

rickecon:
This PR is meant to resolve Issue #770. Currently, a complete run of the unit tests in OG-Core using `pytest` saves output to multiple places in the repository.
```
OUTPUT_BASELINE/
ogcore/tests/OUTPUT_REFORM_/
ogcore/tests/TPI/
ogcore/tests/test_out.pkl
output.csv
output.json
output.tex
output.xlsx
```
This is not a problem for CI runs with GH Actions. However, when running tests on your local machine (despite some of these included in the `.gitignore` file) it is a nuisance to have to manually clean those out of the repository after each testing run on a local machine. This PR seeks to use the `pytest` fixtures `tmpdir` and `tmp_path` to place all saved output from the testing suite into temporary directories.

cc: @jdebacker


rickecon:
@jdebacker. I am currently having trouble getting [`test_run_example.py`](https://github.com/PSLmodels/OG-Core/blob/master/ogcore/tests/test_run_example.py) to save its output to a temporary directory. The [`call_run_ogcore_example()`](https://github.com/PSLmodels/OG-Core/blob/master/ogcore/tests/test_run_example.py#L14) function in lines 14-24 loads the `run_ogcore_example.py` file into the environment as a module. Then the [`test_run_ogcore_example()`](https://github.com/PSLmodels/OG-Core/blob/master/ogcore/tests/test_run_example.py#L28) function uses the `multiprocessing.Process()` function to call the module and run it. The test just makes sure that the script (module) runs for 5 minutes without an error.

This test currently creates saved output from the run in the `./run_examples/` directory.

rickecon:
@jdebacker. Without any changes, I noticed that two tests are failing (all of which are local only) in the `test_run_example.py` script. These tests just make sure the steady-state solves with different user inputs.
```
ogcore/tests/test_user_inputs.py .F.F.....                               [100%]
```
The first failure is in the `test_frisch()` test for a Frisch elasticity of 0.4. The traceback is in the following line from `run_SS`.
```
sol = opt.root(SS_fsolve, guesses, args=ss_params_baseline,
                        method=p.SS_root_method, tol=p.mindist_SS)
if ENFORCE_SOLUTION_CHECKS and not sol.success:
    raise RuntimeError('Steady state equilibrium not found')
E               RuntimeError: Steady state equilibrium not found

ogcore/SS.py:598: RuntimeError
```
The second failure is in the `test_gy()` test for a growth rate of 0.0. The traceback is the same as above.

jdebacker:
@rickecon Thanks for this PR.  The new test failures in testing user inputs is surprising.  It's not obvious why those would not fail - and a bit concerning that a Frisch elasticity of 0.4 fails...

rickecon:
@jdebacker. I have now gone through all the tests. The following three checkbox items are the steps that need to be done for this PR to be ready. I will get back on this PR this weekend or next week. I have to shift gears for the next few days.

- [ ]  `test_run_example.py`. The [`call_run_ogcore_example()`](https://github.com/PSLmodels/OG-Core/blob/master/ogcore/tests/test_run_example.py#L14) function in lines 14-24 loads the `run_ogcore_example.py` file into the environment as a module. Then the [`test_run_ogcore_example()`](https://github.com/PSLmodels/OG-Core/blob/master/ogcore/tests/test_run_example.py#L28) function uses the `multiprocessing.Process()` function to call the module and run it. The test just makes sure that the script (module) runs for 5 minutes without an error. This test currently creates saved output from the run in the `./run_examples/` directory.

- [ ] `test_TPI.py`: all tests with a reform (not baseline) fail because they depend on a file existing from a previous test. The failures are the two `baseline=False` tests in the `test_run_TPI_full_run()` function, the one `baseline=False` test in the `test_run_TPI()` function, and the one `baseline=False` test in the `test_run_TPI_extra()` function. In each case, the error is a FileNotFound in SS.py
```
ogcore/tests/test_TPI.py .........FF.....F.....F.                        [100%]

 >       with open(file_path, 'rb') as f:
E       FileNotFoundError: [Errno 2] No such file or directory: '/private/var/folders/_2/fkr0rfrx2fz2r4_zpbgtdvzw0000gn/T/pytest-of-richardevans/pytest-32/test_run_TPI_full_run_Reform_0/baseline/SS/SS_vars.pkl'

ogcore/utils.py:360: FileNotFoundError
```

- [ ] `test_user_inputs.py`. Two tests are failing in this module, which need to be addressed. As mentioned above, this test module saves no output to disc. However, two tests are failing: Frisch elasticity of 0.4 and productivity growth rate of 0.0. The error in both cases happens in the `run_SS()` function of `SS.py`.
```
ogcore/tests/test_user_inputs.py .F.F.....                               [100%]

sol = opt.root(SS_fsolve, guesses, args=ss_params_baseline,
                        method=p.SS_root_method, tol=p.mindist_SS)
if ENFORCE_SOLUTION_CHECKS and not sol.success:
    raise RuntimeError('Steady state equilibrium not found')
E               RuntimeError: Steady state equilibrium not found

ogcore/SS.py:598: RuntimeError
```

rickecon:
@jdebacker. I fixed the two test failures in `test_user_inputs.py` listed in the previous comment by changing the `initial_guess_r_SS` values in the two offending cases. In the `test_frisch()` test for `frisch=0.4`, I changed `initial_guess_r_SS=0.06` rather than its default value of 0.0648. In the `test_gy()` test for `g_y_annual=0.0`, I also changed `initial_guess_r_SS=0.06` rather than its default value of 0.0648. Both of these changes allowed all the `test_user_inputs.py` tests to pass.
```
ogcore/tests/test_user_inputs.py .........                               [100%]
```

rickecon:
@jdebacker. I fixed all the `reform` test failures in `test_TPI.py` referenced two comments before this one. The two errors in the `test_run_TPI_full_run()` tests were solved by pointing the reform cases to the baseline output already saved in the `./ogcore/tests/test_io_data/OUTPUT/` folder. Those baseline output are for `S=80` and `J=7`.

In contrast, the tests in `test_run_TPI()` and `test_run_TPI_extra()` are based on output for `S=40` and `J=2`. For these, I ran some baseline output and saved the new output in a new folder `./ogcore/tests/test_io_data/OUTPUT2/`.

All the tests in `test_TPI.py` now pass and save all their output to temporary directories.

rickecon:
@jdebacker. This PR is ready for review and merge.

I just finished the last fix to the `test_run_example.py` test. Because it uses the `multiprocessing.Process()` function to call the module, I couldn't figure out a way to call the module from a temporary directory and track its "aliveness" with the `multiprocessing.Process()` function. So instead, I just used the `shutil` package to delete the created directory right before the `assert` statement.

jdebacker:
@rickecon This PR LGTM and I'll merge.  I was worried about the failure of `test_user_inputs.py`, since nothing in this PR should have affected that.  However, I ran those tests locally on different branches and I think the failures came in with PR #767.  That said, we should make a note of the lake of robustness here when running the model with relatively small (and reasonable) changes in the values of important parameters.

rickecon:
@jdebacker. You are right that the failures in `test_user_inputs.py` came in with PR #767. I didn't do the whole set of CI and local tests with that PR because it was much smaller than the previous PR #722. But we know the sensitivity of the root finder. And we also know that a usual fix is to play with the initial guess for `r_SS`.

Issue #781, opened 2022-02-07
Another bug when following run_example directions


martinholmer:
When I try to work around the bug reported in issue #780, I get the following differences from the expected output of the example:

```
(ogcore-dev) run_examples% ls -l *csv
-rw-r--r--@ 1 mrh  staff  1584 Jan 21 15:34 expected_ogusa_example_output.csv
-rw-r--r--  1 mrh  staff  1640 Feb  6 19:56 ogcore_example_output.csv

(ogcore-dev) run_examples% diff ogcore_example_output.csv expected_ogusa_example_output.csv 
1,7c1,7
< ,Variable,2021,2022,2023,2024,2025,2026,2027,2028,2029,2030,2021-2030,SS
< 0,GDP ($Y_t$),-0.16735054602134064,-0.14157894238355173,-0.21744030762126618,-0.2228052844188786,-0.21053121804236263,-0.18645099156078335,-0.16038127595033963,-0.1227400770701889,-0.0966153460839421,-0.0686338595948621,-0.16109691147992355,-0.38614414003818304
< 1,Consumption ($C_t$),-0.0758758816984031,1.0241899555137248,0.12310086751513638,-0.12812369754705397,-0.24351352511095206,-0.32277894213513253,-0.3650822883087033,-0.4194900817886846,-0.4286449823367036,-0.43497670601333205,-0.12405815155773192,-0.7176647471705253
< 2,Capital Stock ($K_t$),-0.37345099560834866,-0.4088108703473829,-0.6526898850168946,-0.7238091470195364,-0.7435953157488909,-0.7329467510562223,-0.7061826074090103,-0.6395066073284759,-0.5915818248341471,-0.5359026282768895,-0.6114964655711641,-1.5067823096305364
< 3,Labor ($L_t$),-0.05611097430623594,0.0026962912189517642,0.0178107834416318,0.048116426639532126,0.07779800378985717,0.10917626595672396,0.13487374038967076,0.1567571028593383,0.17105135740113175,0.18400991222994992,0.08132170577305775,0.22254945389625472
< 4,Real interest rate ($r_t$),-3.0520263383185156,-2.9453224967669973,-2.6454456597459677,-2.528842883299367,-2.472932542916348,-2.458652343235198,-2.460178345199604,-2.587338980604573,-2.616194455888617,-2.658686368570325,-2.645123883468417,-1.4021961225701403
[...]

Issue #780, opened 2022-02-07
Bug when following run_example directions


martinholmer:
The instructions say:
```
The CSV output file `./run_examples/ogcore_example_output.csv` can be
compared to the `./run_examples/expected_ogcore_example_output.csv`
file that is checked into the repository to confirm that you are 
generating the expected output. The easiest way to do this is to use 
the `sh example-diffs` command ...
```

When I try to do that (after running the `python run_ogcore_example.py`
command), I have this unexpected experience:

```
(ogcore-dev) run_examples% ls -l
total 48
drwxr-xr-x   5 mrh  staff   160 Feb  6 18:51 OUTPUT_BASELINE
drwxr-xr-x   5 mrh  staff   160 Feb  6 19:24 OUTPUT_REFORM
drwxr-xr-x   4 mrh  staff   128 Feb  6 19:56 dask-worker-space
-rw-r--r--@  1 mrh  staff    81 Jan 21 15:34 example-diffs
-rw-r--r--@  1 mrh  staff    64 Jan 21 15:34 example-diffs.bat
-rw-r--r--@  1 mrh  staff  1584 Jan 21 15:34 expected_ogusa_example_output.csv
-rw-r--r--   1 mrh  staff  1640 Feb  6 19:56 ogcore_example_output.csv
drwxr-xr-x  33 mrh  staff  1056 Feb  6 19:56 run_example_plots
-rw-r--r--@  1 mrh  staff  4271 Jan 21 15:34 run_ogcore_example.py

(ogcore-dev) run_examples% sh example-diffs
diff: expected_ogcore_example_output.csv: No such file or directory
```
Looks like you have forgotten to rename the `expected_ogusa_example_output.csv` file.

rickecon:
This issue is resolved by PR #785. Closing.

PR #785, opened 2022-02-10
Add test of output from run_example.py

jdebacker:
This PR adds a unit test of the output from `run_example.py`.  It also updates the expected output from that script that one can test against with their own local run of the script.

Addresses Issues #780 and #781


rickecon:
@jdebacker. It looks like just three tests are failing in `test_parameter_plots.py`.
```
ogcore/tests/test_parameter_plots.py ...............................FFF. [ 64%]
```
The three failures are the same error in the following three tests: `test_plot_2D_taxfunc[with data]`, `test_plot_2D_taxfunc[MTR capital income] `, and `test_plot_2D_taxfunc[MTR labor income]`. The error is in the call to the `parameter_plots.py` module in the `wm = lambda x: np.average()` function on line 975 of the `plot_2D_taxfunc()` function. The error is `ZeroDivisionError: Weights sum to zero, can't be normalized`.

jdebacker:
@rickecon, This PR is ready for your review.

PR #790, opened 2022-03-08
Update setup.py

jdebacker:
This PR makes a few updates to `setup.py` necessary for publishing `ogcore` to PyPI:
* Change wording for license description
* Note that long description is being read from a md file
* Update version number for consistency with repo

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,15 +6,16 @@ except ImportError:
 with open('README.md') as f:
     longdesc = f.read()
 
-version = '0.7.0'
+version = '0.8.0'
 
 config = {
-    'description': 'General equilibribum, overlapping generations model for the USA',
+    'description': 'A general equilibribum overlapping generations model for fiscal policy analysis',
+    'long_description_content_type': 'text/markdown',
     'long_description': longdesc,
     'url': 'https://github.com/PSLmodels/OG-Core/',
     'download_url': 'https://github.com/PLSmodels/OG-Core/',
     'version': version,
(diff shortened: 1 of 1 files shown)


rickecon:
@jdebacker. Is this ready to merge once all the checks pass? It looks good to me.

jdebacker:
@rickecon I still have one outstanding issue, though I'm not sure it can be resolved.  Here's my comment from Element:

>Created a test package of OG-Core: https://test.pypi.org/project/ogcore/0.8.0/ but it doesn't seem to install the mkl dependency. I needed to do that separately then package installed ok. Anyone else having this issue? Thoughts on why?

rickecon:
@jdebacker. Two comments

1. Potentially it might be helpful to list Python 3.10 as a programming language after line 42 in `setup.py`. There might be some issue where the packages needed require 3.10 for all of them to work. (This is likely not the case, but we should have this line anyway.
```
'Programming Language :: Python :: 3.10',
```

2. Despite [`mkl` being on PyPI](https://pypi.org/project/mkl/), a lot of the online forums suggest downloading through `conda`. (see [here](https://stackoverflow.com/questions/70247399/error-no-matching-distribution-found-for-mkl-service-2-3-0-gcp-ec2) and [here](https://stackoverflow.com/questions/62688075/python-pip-install-mkl-fft-could-not-install))

But this really doesn't make any sense. I think the best way to do this is to use pypi.org so the package can be pip installed. It would be really nice to get this to work.

PR #810, opened 2022-07-20
Add investment tax credit

jdebacker:
This PR adds an investment tax credit parameter to the model.

PR #814, opened 2022-08-16
Bug fix for investment tax credit

jdebacker:
This PR fixes a bug introduced in PR #810.  In that PR, the investment tax credit was modeled realistically (where the credit rate is applied to the amount of investment), but because the firms in OG-Core don't make investment decisions (they just rent capital each period, not choosing the amount of the capital for the next), this led to an error in how the investment tax credit showed up in the firms' FOCs.  The result was that market clearing was not satisfied.

This PR fixes the bug by modeling the investment tax credit as being applied to a proxy for the amount of investment, delta * K.  This matches investment in the SS.  While it does introduce some approximation error in the transition path, it does help to capture the incentive effects that investment tax credits present - lowering the cost of capital.


rickecon:
@jdebacker. Will you please do the following two things to this PR.

* Run `black` on the three files (or all files) highlighted as failing in the check black CI tests.
* Check to see why the test is failing in `test_utils.py` in the `test_read_file_from_egg()` function giving a `FileNotFoundError` for missing `default_parameters.json`. I have just re-run those tests, but I don't think they will pass.
```
tests/test_utils.py .....F.............................................. [ 97%]

___________________________ test_read_file_from_egg ____________________________

    def test_read_file_from_egg():
        """
        Test of utils.read_file() function, case of reading file from .egg
        """
        path = os.path.join(CUR_PATH)
        fname = "default_parameters.json"
>       bytes_data = utils.read_file(path, fname)

tests/test_utils.py:78: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
ogcore/utils.py:93: in read_file
    buf = resource_stream(Requirement.parse("ogcore"), path_in_egg)
/usr/share/miniconda/envs/ogcore-dev/lib/python3.9/site-packages/pkg_resources/__init__.py:1160: in resource_stream
    return get_provider(package_or_requirement).get_resource_stream(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pkg_resources.PathMetadata object at 0x7fa[33](https://github.com/PSLmodels/OG-Core/runs/7860796552?check_suite_focus=true#step:5:34)9bb8fa0>
manager = <pkg_resources.ResourceManager object at 0x7fa3[39](https://github.com/PSLmodels/OG-Core/runs/7860796552?check_suite_focus=true#step:5:40)c1d280>
resource_name = 'ogcore/default_parameters.json'

    def get_resource_stream(self, manager, resource_name):
>       return open(self._fn(self.module_path, resource_name), 'rb')
E       FileNotFoundError: [Errno 2] No such file or directory: '/usr/share/miniconda/envs/ogcore-dev/lib/python3.9/site-packages/ogcore/default_parameters.json'
```

jdebacker:
@rickecon I've updated the formatting.

I also ran `test_utils.py` locally and get that same failure.  As far as I can tell, the `utils.read_file` function tested here is not used anywhere in the code base.  I can't explain the failure, but maybe this function could be removed?

rickecon:
@jdebacker. I checked as well, and the `utils.read_file()` function is not used anywhere in the code. That means you can do the following:
* Delete the `read_file()` function in `utils.py` (lines 77-97)
* Remove `read_file from the `docs/book/content/api/utils.rst` file members (line 17)
* Delete two tests `` and `` from `test_utils.py` (lines 61-80)

It is a weird question why this error would start showing up now.

jdebacker:
@rickecon Actually I know see that the `utils.read_file()` function *is* used in the country-specific models (e.g., OG-USA).  So we should keep it. Let me look more into why this might be an issue now when it wasn't with PRs a few weeks ago.

rickecon:
@jdebacker. But if the function is not used in `OG-Core`, then I think we should move the function to a `utils` module in the country calibration repos and do the testing there.

jdebacker:
@rickecon writes:

> But if the function is not used in OG-Core, then I think we should move the function to a utils module in the country calibration repos and do the testing there.

I think the exception here is if the function is used in every country specific model.  e.g., we only have `utils.read_cbo_forecast()` in the OG-USA repo.  And most of the functions in `txfunc.py` are not called in OG-Core, but because they are used in every country-specific model's calibration, we've kept them in OG-Core. This reduced duplicate code and makes maintenance easier.

jdebacker:
FYI, all tests now passing locally.  At first, the two tests in `tests/test_run_example.py` didn't pass (there was an error with finding the `OUTPUT_BASELINE` directory), but when running them again (with no changes), they passed.

rickecon:
@jdebacker . I just submitted [a PR](https://github.com/jdebacker/OG-Core/pull/28) to your branch for this PR that has some documentation updates. Once that is merged in, this PR is ready to merge. Thanks.

jdebacker:
@rickecon I just reviewed and merged your PR to my branch.  Thanks for those updates!

rickecon:
@jdebacker. Now everything is good to merge. I did have to re-run the Python 3.10 build and test. It still perplexes me why that fails sometimes and passes other times.

PR #816, opened 2022-08-29
Modeling income tax compliance

jdebacker:
This PR adds two new parameters (`labor_income_tax_noncompliance_rate` and `capital_income_tax_noncompliance_rate`).  These will allow the OG-Core model to account for the revenue and incentive effects of income tax noncompliance.  Note that this is important since the tax function parameter are typically estimated using data that miss the effects of compliance (either because they use administrative data from tax returns or because of similar under-reporting in survey data (see [Hurst et al. (_REStat_, 2014)](https://direct.mit.edu/rest/article-abstract/96/1/19/58117/Are-Household-Surveys-Like-Tax-Forms-Evidence-from))). 

This PR is a work in progress until the following items are addressed:
- [x] test the `tax.MTR_income` and `tax.ETR_income` functions with non-zero noncompliance
- [x] test `tax.net_taxes` function (which calls `tax.income_tax_liab`, fo which there are no separate tests) with non-zero noncompliance for the following combinations of parameterizations:
  - [X] SS, `j!=None`
  - [ X] TPI, `j!=None`
- [x] test `household.FOC_savings` function with non-zero noncompliance for the following combinations of parameterizations:
  - [x] SS, `j = None`
  - [x] SS, `j!=None`
  - [x] ~~TPI, `j=None`~~ This case is never used in model
  - [x] TPI, `j!=None`
- [x] test `household.FOC_labor` function with non-zero noncompliance for the following combinations of parameterizations:
  - [x] SS, `j = None`
  - [x] SS, `j!=None`
  - [x] ~~TPI, `j=None`~~ This case is never used in model
  - [x] TPI, `j!=None`
- [x] The tax noncompliance parameters are described in the documentation

PR #818, opened 2022-09-12
Use IO-Matrix to map production goods to consumption goods

jdebacker:
This PR adds and IO-matrix parameter to map production goods into consumption goods.

WIP will be removed after the following items are addressed:
- [x] update documentation to reflect the use of the IO matrix
- [x] add test for `SS.inner_loop` with `I, M` > 2` & `I != M`
- [x] add test for TPI.inner_loop with `I, M > 2` & `I != M`


jdebacker:
The above items have been addressed, but per conversations with @rickecon I will add the following to this PR:

- [ ] A vector representing a fixed-coefficient model of government goods and services (where each element is the fraction of a unit of gov't expenditures that come from each industry `m`)
- [ ] Tests of the above

At this point, I do not have plans to similarly model the components of net exports, which right now are assumed to come fully from the capital producing industry.

jdebacker:

rickecon:
@jdebacker. Can you resolve the `parameters.py` conflicts coming from the merged noncompliance PR #816? These look like they are nontrivial to me.

jdebacker:
Looks like this branch is now getting test failures due to a change in the Matplotlib API for 3d plots.  That is fixed over in PR #819, so we'll merge that to master first then update this branch, which should take care of the test failures.

All tests now passing for me on local runs.

jdebacker:
@rickecon This should be ready to go after your review.

ogcore/default_parameters.json
```diff
@@ -906,7 +947,7 @@
         "description": "Consumption tax rate.  Set value for base year, click '+' to add value for next year.  All future years not specified are set to last value entered.",
         "section_1": "Fiscal Policy Parameters",
         "section_2": "Taxes",
-        "notes": "This policy parameter represents the effective consumption tax rate from sales taxes, VATs, and excise taxes.  To capture exemptions of certain goods, this is assumed to vary by age and ability type.  It is thus a SxJ matrix.",
+        "notes": "This policy parameter represents the effective consumption tax rate from sales taxes, VATs, and excise taxes.  To capture exemptions of certain goods, this is assumed to vary by age and ability type.  It is thus a IxSxJ array.",
```
rickecon, line 950: @jdebacker. Given that `tau_c` is now `I x S x J`, shouldn't the default value have one more bracket?

rickecon:
@jdebacker. In the Python 3.9 CI tests have failed twice in the same place--the first test in `test_SS.py`. Do you want me to look into this?

rickecon:
@jdebacker. I think I have isolated the problem. The error goes away if I set the `environment.yml` requirement for SciPy to `scipy>=1.8.1`. However, this requires the additional environment setting of `python>=3.8`. The current Python and SciPy environment settings are `python>=3.7.7` and `scipy>=1.7.1`. The current SciPy lower bound of `1.7.1` allows us to use Python 3.7.7 and maybe some of the other requirements that interact with the Python 3.7.7 requirement.

I did some testing of other environment configurations.

1) `python=3.7.7` and `scipy>=1.8.1`. This broke. Conda could not find a configuration of required packages that were compatible.

2) `python=3.8` and `scipy>=1.8.1`. Worked great and did not break.

3) `python=3.9` and `scipy>=1.8.1`. Worked great, did not break, and all `test_SS.py` tests passed.

**CONCLUSION:** The jump from SciPy 1.7.1 to 1.8.1 is significant. The results of the root finder change. And v. 1.8.1 is not compatible with Python 3.7.7 and the other requirements of the package. We should ditch Python 3.7 (`python>=3.8`) and commit to Scipy greater than or equal to 1.8.1 (`scipy>=1.8.1`).

I have submitted this change in a [PR to your branch](https://github.com/jdebacker/OG-Core/pull/31). If you merge that in, I predict our CI error in Python 3.9 will go away. But I don't know for sure. I only know that the error is not present on my machine when I performed scenario (3) above.

rickecon:
@jdebacker. OK. I am hiring a CS student from Princeton, and one of his jobs is to refactor `openfisca-core` to be compliant with Python 3.8+. But I guess for now we can just ignore these test failures in `test_SS.py`. If you are OK with that, then I am ready to merge this PR.

rickecon:
@jdebacker. We are merging this PR in spite of the CI failure in the Python 3.9 `test_SS.py` test. We have isolated this test failure to a conflict between versions of SciPy and the requirement to be compatible with Python 3.7.7. This problem will go away once the openfisca requirement for Python 3.7 is relaxed.

Issue #823, opened 2022-10-28
Allow for flexible data years in plot_population_path


jdebacker:
The function `parameter_plots.plot_population_path()` needs to be updated to allow for labels that reflect the data and model start years.

jdebacker:
Resolved with PR #825

PR #825, opened 2022-11-09
Remove hard coded year in plot of population distribution over the time path

jdebacker:
This PR addresses Issue #823 by removing the hard coded label for the year used for the initial population in the `parameter_plots.plot_population_path()` function.


rickecon:
@jdebacker. This looks great. Merging.

PR #836, opened 2022-12-20
Update build_and_test.yml GitHub Action

rickecon:
This PR updates the `build_and_test.yml` in the following ways:
* Adds three operating systems, including `Mac OS` to `build_and_test.yml`
* Removes the Python 3.8 tests, so this only tests Python 3.9 and 3.10
* Updates to `actions/checkout@v3`
* Updates to `codecov/codecov-action@v3`
* Remove unused `from curses.ascii import SP` from `test_firms.py`. This unused import was causing the tests to fail on Windows latest platform with Python 3.9.
* Update and add badges to the `README.md`
* Updated expected values and tolerance level in `test_SS_fsolve()` test and updated tolerance level in `test_run_SS()` test in `test_SS.py`
* Added tolerance level to `test_twist_doughnut()` in test_TPI.py`

PR #838, opened 2022-12-21
Update max values for r_gov_scale and r_gov_shift in default_parameters.json

rickecon:
This PR updates the maximum values for `r_gov_scale` and `r_gov_shift` from 2.0 and 0.2, respectively, to 4.0 and 0.3. This is because the calibrated values of these two parameters for the OG-ZAF (South Africa) repository are respectively, 3.37662504 and 0.24484764, which are slightly above the current maxima for these parameters (See [PR #24](https://github.com/EAPD-DRB/OG-ZAF/pull/24) in OG-ZAF by @SeaCelo).

I also updated the references to `np.float()` to be `np.float64` in `parameter_plots.py`. The lack of this change was causing tests to fail in `test_parameter_plots.py`.

cc: @jdebacker

### DIFF
--- a/ogcore/default_parameters.json
+++ b/ogcore/default_parameters.json
@@ -837,7 +837,7 @@
         "validators": {
             "range": {
                 "min": 0.0,
-                "max": 2.0
+                "max": 4.0
             }
         }
     },
@@ -855,7 +855,7 @@
         "validators": {
             "range": {
                 "min": 0.0,
-                "max": 0.2
+                "max": 0.3
             }
         }
     },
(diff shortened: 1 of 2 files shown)


rickecon:
@jdebacker. This PR is ready for your review as soon as you have time.

rickecon:
@jdebacker. I recommend merging this PR in spite of it not passing the `test_SS_fsolve()` test in `test_SS.py`. I am trying to fix that in PR #836. All this PR is doing is expanding the maxima for `r_gov_shift` and `r_gov_scale`.

PR #845, opened 2023-01-12
Bump the version to include recent PR in PyPI package

rickecon:
@jdebacker. This PR updates the version of OG-Core so that the most recent PR can be included in the PyPI.org package. This PR updates the version number to 0.10.2 in `setup.py` and `__init__.py`. I also made a small change in the Python versions referenced in `contributor_guide.md`.


jdebacker:
Thanks for the version bump, @rickecon.

Issue #16, opened 2015-04-03
Switch to CFE disultility approximation for elliptical utility.


kerkphil:
What value of the Frisch elasticity to we want to use for the approximation?

kerkphil:
Peterman (2014) notes:
"These results demonstrate that the value used to calibrate the Frisch elasticity in
a macroeconomic model depends crucially on both the question the economist is
asking and the specific features of the model he is including. For example, if a
macroeconomist is using a model that does not include retirement and it is important
for answering his question that the 
fluctuations in hours and wages over the business
cycle are consistent with the data then he will need to use a value consistent in
line with the large macro Frisch elasticity estimates in this paper. In contrast, if an
economist is asking a question that centers on changes in hours on the intensive margin
of working individuals then he may want to use a lower calibration value for the Frisch
elasticity in line with the estimates of the micro Frisch elasticity."

kerkphil:
Uploaded papers by Peterman (2014) and Keane & Rogerson (2012) on the Frisch elasticity.

kerkphil:
With a Frisch elasticity 0f 1.5 the ellipse parameters are: b = 0.599331969 and mu = 1.360744323
If we use .9 we get b = 0.43241947 and  = 1.543691436

jdebacker:
With PR #178 we have added the ability for the user to enter a Frisch elasticity and have the elliptical utility parameters fit to a constant Frisch elasticity function with the given parameterization.

PR #846, opened 2023-01-20
Fix conditional expression in txfunc.py

jdebacker:
This PR fixes a bug in a conditional expression in `txfunc.py`.  Given the multiple conditions and the strict equality condition, conditions were not evaluated in the expected way (see [this Stack Overflow thread](https://stackoverflow.com/questions/50656307/numpy-typeerror-ufunc-bitwise-and-not-supported-for-the-input-types-when-us) on the precedence of `&` and `==`).


jdebacker:
@rickecon This PR is ready for review.  Without the change, integration of the microsimulation model in [OG-IND in PR #16](https://github.com/Revenue-Academy/OG-IND/pull/16) fails when it reaches this condition.

rickecon:
@jdebacker. Looks good to me.

PR #839, opened 2022-12-27
Generalize spline tax functions to 2D

prrathi:
This PR addresses issue #828 through [pyGAM](https://pygam.readthedocs.io/en/latest/) which enables monotonic splines for any number of dimensions. It creates a new method within `monotone_spline` function alongside new arguments to specify the method and necessary inputs. Also adds a new function `avg_by_bin_2d` for binning data with any number of x dimensions (maybe the function name should be changed) that's used in the 2d case for the tax data example created in `test_txfunc.py`. @jdebacker @rickecon

PR #848, opened 2023-01-21
Release 0.10.3

jdebacker:
This PR prepares the release of OG-Core v 0.10.3

### DIFF
--- a/ogcore/__init__.py
+++ b/ogcore/__init__.py
@@ -19,4 +19,4 @@ from ogcore.tax import *
 from ogcore.txfunc import *
 from ogcore.utils import *
 
-__version__ = "0.10.2"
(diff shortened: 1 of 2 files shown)


rickecon:
@jdebacker. Do you want to wait to bump the version until we get PR #839 merged? The only PR since version 0.10.2 is PR #846. I just met with Praneet last night and I have everything I need to review his 2D monotonic tax functions. They are really good, and I think that would be the right time to bump the version.

jdebacker:
@rickecon This has an important bug fix for work I'd like to complete in OG-IND, so I want to do a minor version release and publish a new package on PyPI.  We can do another after the tax function PR is reviewed and merged too.

PR #857, opened 2023-02-15
Updates to plotting functions

jdebacker:
This PR makes the following updates:

* Use 300 dpi when saving plots to disk
* Better labels of the `plot_industry_aggregates` plots


rickecon:
@jdebacker. Looks great. Merging.

PR #457, opened 2019-04-20
[WIP] Dynamic Firms

jdebacker:
This PR replaces the static representative firm with a dynamic representative firm, that maximize the present discounted value of profits.  This addition allows for modeling capital adjustment costs and better modeling of business tax policies.

Issue #406, opened 2018-10-22
Capital adjustment costs


jdebacker:
This issue is to discuss the best approach to add capital adjustment costs to the model.  

One way to do this is to have firms own capital and households holding equity in these firms.  This will change the firm FOC for capital and the interpretation of household savings.  It will also change the market clearing conditions capital and savings (namely, the model will then have an asset clearing condition that total household savings equals the total value of firm equity).

jdebacker:
A roadmap for this model enhancement:

- [ ] Write functions for adjustment costs and their derivatives
- [ ] (1) Write `firm.get_LY_ratio` analogous to the `firm.get_KY_ratio` function, but to solve for the labor output ratio rather than the capital output ratio
- [ ] (2) Write `firm.solve_K` analogous to the `firm.solve_L` function, but to solve for `K` from the production function given `Y` and `L`
- [ ] Update the `firm.get_r` function to include (the derivatives of) the adjustment cost function
- [ ] Update solution algorithm to put in place (1) and (2) and a computation of total adjustment costs by industry to pass to the resource constraint function (I suppose these are all paid in units of the numeraire industry output?)
- [ ] Write a function for dividends (= after tax cash flow of firm - need not be positive (negative is equity issuance))
- [ ] Update `aggr.get_r_p` to include dividends in place of `rK`
- [ ] Update docs to reflect these new functions and solution algorithm

Issue #332, opened 2017-12-06
How to treat tax basis of capital stock


rickecon:
In OG-USA right now, the firm's profit function is the following,

![proffunc](https://user-images.githubusercontent.com/4317690/33684558-48eed68a-da94-11e7-8732-813576acf666.png)

where `delta**tau` in the last term is the tax basis of depreciation. It represents the fraction of economic depreciation that can be deducted from the firm's tax liability at the corporate income tax rate `tau**corp`. One issue with this is that when the tax basis depreciation is not equal to the economic depreciation `tau**delta != delta` then the tax basis of aggregate capital does not equal the actual aggregate capital `K**tau != K`. The last term in the profit equation should really be in terms of the tax basis of aggregate capital `K**tau`.

![proffunctau](https://user-images.githubusercontent.com/4317690/33684764-0081cd66-da95-11e7-9c61-316c3b545935.png)

Making the tax basis of aggregate capital the component that determines the degree of depreciation deductibility makes `K**tau` the component that firms need to keep track of. Intuitively, the tax basis of aggregate capital is a function of the entire history of actual aggregate capital stocks. This is seen by combining the law of motion for the actual aggregate capital stock,

![klom](https://user-images.githubusercontent.com/4317690/33685026-e3780112-da95-11e7-95f7-8acd81d82787.png)

with the law of motion for the tax basis of the aggregate capital stock.

![ktaulom](https://user-images.githubusercontent.com/4317690/33685069-03bd3b86-da96-11e7-86a0-7a204abd0ec3.png)

Subtracting the actual law of motion equation from the tax basis law of motion equation and solving for `K**tau_{t+1}` gives the recursion that shows that the tax basis aggregate capital is a function of the history of actual capital stocks.

![ktautp1](https://user-images.githubusercontent.com/4317690/33685158-49c8d054-da96-11e7-928f-af455c7cef34.png)

This renders the firm's problem dynamic. We need to update this in OG-USA.

PR #874, opened 2023-08-11
Dynamic firms across multiple industries

jdebacker:
This PR changes the model to allow for firms that own their own capital and face capital adjustment costs, but who also differ in the production goods they produce. 

This PR will address Issue #406 and #332.  It will replace PR #457.


jdebacker:
@rickecon I keep getting a failure in the code formatting test.  But when I run `black` locally, it says no changes needed:
```
(ogcore-dev) jason.debacker@JDEBACKER-7 OG-Core % black -l 79 ./
Skipping .ipynb files as Jupyter dependencies are not installed.
You can fix this by running ``pip install "black[jupyter]"``
All done! ✨ 🍰 ✨
45 files left unchanged.
```
I am using `black` version 23.11.0, which is the latest version. Any ideas why this test isn't passing?

vahid-ahmadi:
Draft, so treating this as directional. The economics of endogenising the cost of capital through the adjustment-cost derivatives looks right and the theory files carry enough of the derivation to follow.

Two things in `get_pm` that will bite when you pick this back up. `get_cost_of_capital = get_cost_of_capital(r, p, method, m)` shadows the function with its own return value, and calls it on the old signature with an `r` that isn't in scope — so the new pricing path can't currently execute. In the same call, `get_MPx(..., 1 - p.gamma[m] - p.gamma_g[m], ...)` is passed alongside `K`, but `share` is documented as "share of output paid to factor x" (`firm.py`), so with `K` that should be `p.gamma[m]` — as written it's labour's share against the capital input.

Your own `# Append what to the front of this??` on `tau_b_m1` is probably the thing to settle first, since the t-1 tax indexing at the head of the path constrains the rest. It still merges clean, so this is dormant rather than blocked.

Issue #877, opened 2023-08-24
Coverage of various dimensions for tax parameter inputs


jdebacker:
Lines 351 to 398 of `parameters.py` attempt to make the tax parameters conform with the model dimensions for those parameters (TxSxnum params) from variously shaped inputs (e.g., if one gives a single number, it's assumed they want constant rates across model periods and ages, or if a list of length S is input, it is assumed they want age varying tax rates that are constant across time periods).  This all got more difficulty when moving the parameter contain from Numpy arrays to lists and there is an if statement that is missing that extends along the age dimension if it's not equal to S.

This has caused issues, such as [this one](https://github.com/EAPD-DRB/OG-ZAF/issues/43) in the OG-ZAF project.

jdebacker:
Resolved with merge of PR #878

PR #878, opened 2023-09-01
Fix tax function indexing

jdebacker:
This PR fixes two cases of tax function parameters not being indexed (or sized) correctly.

The first is a change to `parameters.py` to make sure tax functions that are input are extrapolated to the appropriate model dimensions.  This addresses Issue #877.

Second is a changes to the plotting of 2D tax functions in `parameter_plots.py`. This function had not been updated (or not correctly updated) to notes that the dimensions of the lists of lists of tax functions parameters are TxS. The change here is to make sure time period index is given before the age index.


jdebacker:
All unit tests are passing and I've run the example script successfully my my machine.

There is some issue building miniconda in the "Check that docs build" workflow. It continues to time out, even after re-running this job 3 times. I do not know the issue since the same actions are used to build miniconda here as are used in the successful unit test actions.

If @rickecon doesn't have a quick solution, I would say go ahead and merge this PR. It is otherwise ready to go.

environment.yml
```diff
@@ -2,7 +2,7 @@ name: ogcore-dev
 channels:
 - conda-forge
 dependencies:
-- python>=3.7.7, <3.11  # This restriction can be removed as soon as we test Python 3.11
+- python>=3.7.7, <=3.11  # This restriction can be removed as soon as we test Python 3.11
```
rickecon, line 5: @jdebacker. You can get rid of this comment at the end of this line.

rickecon:
@jdebacker. All tests have passed except for `Build and test [Python 3.9, 3.10] / build (ubuntu-latest, 3.9)`. Maybe we need to use `mambaforge` for our `build_and_test.yaml`.
OGcoreScreenshot

setup.py
```diff
@@ -23,7 +23,7 @@
         ]
     },
     include_packages=True,
-    python_requires=">=3.7.7, <3.11",
+    python_requires=">=3.7.7, <=3.11",
```
rickecon, line 26: @jdebacker. Parallel to the change in `environment.yml` you should probably remove the `<=3.11` here in `setup.py`.

rickecon:
@jdebacker. We'll see how this goes. I am worried that we will run out of compute time with 9 `build_and_test.yml` matrix runs. If it times out, we can just remove Python 3.9 from the tests. Unfortunately, I don't know an easy way to do "local only" Python 3.11 tests.

rickecon:
@jdebacker. It looks like Python 3.11 is resulting in some `pickle.load` errors. Below is the traceback from the Build and test Linux Python 3.11 tests.
```
ERROR tests/test_TPI.py - _pickle.UnpicklingError: Memo value not found at index 80
ERROR tests/test_output_plots.py - _pickle.UnpicklingError: Memo value not found at index 15776
ERROR tests/test_output_tables.py - _pickle.UnpicklingError: Memo value not found at index 15776
ERROR tests/test_parameter_plots.py - _pickle.UnpicklingError: Memo value not found at index 15776
ERROR tests/test_parameter_tables.py - _pickle.UnpicklingError: Memo value not found at index 15776
```

rickecon:
@jdebacker. The good news is that it ran really fast.

jdebacker:
@rickecon writes:
>@jdebacker. It looks like Python 3.11 is resulting in some pickle.load errors. Below is the traceback from the Build and test Linux Python 3.11 tests.

Yes, if you save a pickle file in an earlier version of Python (or relevant packages that the pickled objects are from), it's difficult to open in a later version, as far as I know.  We've run into this before.  Since I don't want to save all pickle files again right now, I've just removed 3.11 from the test matrix.

rickecon (approved):

rickecon:
Thanks @jdebacker. Looks great. Merging.

rickecon:
Oops. We just need to update the version number in `setup.py` and `ogcore/__init__.py`. Probably bump it up to 0.10.8 rather than 0.11.0. Right?

jdebacker:
@rickecon Let me do a quick PR to resolve Issue #879 before we release a new version.  We've got several changes we can package together. But yet, that will be 0.10.8.

Issue #879, opened 2023-09-05
Make dimensions from TPI output a consistent size


jdebacker:
Some of the model output from the time path solution has `T` periods, while other output variables have `T+S` periods.  It would be helpful to consumers of the model output if the time dimension had a standard length and `T` is probably sufficient (since while we create `T+S` length arrays for the model solution algorithm, the model hits it's steady state by before period `T`).

jdebacker:
Resolved with PR #880

Issue #321, opened 2017-10-25
OG-USA new release after 0.5.9?


hdoupe:
Do we want to release a new version of OG-USA for [TaxBrain reinstatement](https://github.com/OpenSourcePolicyCenter/PolicyBrain/pull/710)?  

I think that this release would include PR's #305, #308, #311, #312, #313, and #315.  One more PR may be necessary to set the path to the PUF file or to take the PUF dataframe as an argument in `execute.runner` and pass that to Tax-Calculator in `get_micro_data.get_calculator`.

@PeterDSteinberg @rickecon @PeterDSteinberg

PeterDSteinberg:
I am not as familiar with the PRs you listed above @hdoupe .  We are also currently working on OG-USA related changes for webapp's issues https://github.com/OpenSourcePolicyCenter/PolicyBrain/issues/481 and https://github.com/OpenSourcePolicyCenter/PolicyBrain/issues/482 . We hope to wrap those new features up this week for 481 and 482.  Maybe deploy after those are addressed?  Or two deploys?

hdoupe:
Ok, I think we are going to be pretty tight on time if we try to include OpenSourcePolicyCenter/PolicyBrain#481 and OpenSourcePolicyCenter/PolicyBrain#482 in the next release on Friday.  We are on a pretty tight release schedule through the rest of the year.  A CPS option and UBI parameters need to be added by the end of the year.  So, I'm wary of two things:

1. Rushing these features in on Friday and running into bugs 
2. Waiting to release until after these features have been merged and then falling behind on our release schedule

It seems to me that the best course of action is to release without OG-USA on Friday and to incorporate it in the next release.  This means that it would not be added for about 3 weeks.

@PeterDSteinberg @jdebacker @rickecon @brittainhard @MattHJensen  what are your thoughts?

PeterDSteinberg:
@hdoupe I was just about to email you.  I spoke with @brittainhard and agree we should not try to get 481 and 482 fixed by Friday.  Also Brittain and I need to rebase the work we have done for 481 on your https://github.com/OpenSourcePolicyCenter/PolicyBrain/pull/710 .  We can go ahead and branch off that Policybrain PR 710 then when 710 is merged we can pull master into our fixes for 481, 482.  That way we can get it ready for merge by Friday but have 3 weeks for testing on test app.

hdoupe:
I think this has been resolved. Closing #321

PR #880, opened 2023-09-07
Standardize time path output

jdebacker:
This PR standardizes the variables saved to the output dictionary from the time path solution.  Previously, some variables had a dimension with length `T+S` while others had a length of `T` in the same dimension (as noted in Issue #879).  This PR sets the first dimension of all time path output variables to be of length `T`.  This consistency in the length along this dimension will help users more easily use the model output.


jdebacker:
This PR is ready to go.  I ran the full suite of tests locally and found 3 failures:

```
FAILED tests/test_txfunc.py::test_monotone_spline - AttributeError: module 'numpy' has no attribute 'int'.
```

The first two are expected to fail since I'm using a different version of SciPy than was used for the cached tax function estimates are checking against (and we know different versions produce slightly different parameter estimates).  The third failure is because the PyGAM package is incompatible with more recent NumPy versions.  If that project would merge [PR #321](https://github.com/dswah/pyGAM/pull/321) (which changes just two lines), this test would pass.

rickecon (approved):

rickecon:
@jdebacker. This looks great. Will you please update the version number to 0.10.8 in `setup.py` and in `ogcore/__init__.py`. As soon as this is done, I will merge.

rickecon:
@jdebacker. BTW, I keep looking for alternatives to [`PyGAM`](https://github.com/dswah/pyGAM) for generalized additive models. It is not clear to me if there is a good alternative in `scikit-learn`. This Jupyter Book "[Generalized Additive Models (GAM)](https://kirenz.github.io/regression/docs/gam.html)" chapter uses the `gam` module of the `statsmodels` package ([`statsmodels.gam`](https://www.statsmodels.org/devel/gam.html)), which might be an alternative.

jdebacker:
@rickecon Version bump done.  Will look into the alternative to PyGAM - thanks.

rickecon (approved):

PR #895, opened 2024-01-16
Time varying ability matrix

jdebacker:
This PR allows one to use an ability/labor productivity matrix that varies over time.  Changing the shape of the `e` parameter object from SxJ to TxSxJ.


run_examples/run_ogcore_example.py
```diff
@@ -73,6 +73,8 @@ def main():
     )
     # Update parameters for baseline from default json file
     p.update_specifications(og_spec)
+    # make e matrix 3D
```
rickecon, line 76: @jdebacker. Shouldn't we make the default e matrix 3-d in `default_parameters.json` lines 170-278? We should change the `number_dims` to 3 in line 177. And we should add one more set of brackets around the input `value` to make it a three dimensional nested list. Then we just need to code `parameters.py` to copy any two dimensional input down the t-dimension--as would be the case in the default specification. What do you think?

rickecon (approved):

rickecon:
@jdebacker. I deleted my `ogcore-dev` conda environment and reinstalled it in my clone of your branch. When I ran `run_ogcore_example.py` and got the following error:
```
(ogcore-dev) richardevans@Richards-MacBook-Pro-2 run_examples % python run_ogcore_example.py
Number of workers =  7
Traceback (most recent call last):
  File "/Users/richardevans/Docs/Economics/OSE/OG-Core/run_examples/run_ogcore_example.py", line 138, in <module>
    main()
  File "/Users/richardevans/Docs/Economics/OSE/OG-Core/run_examples/run_ogcore_example.py", line 77, in main
    p.e = np.tile(p.e.reshape(1, p.S, p.J), (p.T, 1, 1))
                  ^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: cannot reshape array of size 224000 into shape (1,80,7)
```

run_examples/run_ogcore_example.py
jdebacker: @rickecon At some point, we might.  But my thought is to keep this "under the hood" for now.  If we change `default_parameters.json` then we need to update each country repo with the 3D `e`.  

Since we're not sure how many use cases there will be of this, I would say we leave `default_parameters.json` as is, then when one wants to update to a 3D `e`, they just do so outside of the `update_specfications` method (i.e., they just assign `p.e = new_array` with a `new_array` that is of shape `(p.T, p.S, p.J)`.

jdebacker:
@rickecon Noted about the example.  That was for some testing and should be taken out.  Will update.

rickecon (approved):

jdebacker:
All test we expect to pass on Python 3.11 pass locally.  The three failures are the `txfunc.py` functions whose results vary with the version on Python/SciPy used:

```
FAILED tests/test_txfunc.py::test_tax_func_estimate - assert False
========= 3 failed, 496 passed, 16819 warnings in 32781.69s (9:06:21) ==========
```

I was also able to run the `run_ogcore_example.py` script all the way through without issue.

I believe this PR is ready to merge.

rickecon:
LGTM. Thanks, @jdebacker.

PR #896, opened 2024-01-16
Create a demographics module in OG-Core

jdebacker:
This PR creates a demographic module in OG-Core.  Such a module exists in the various country calibration repositories, but since all of those now use the UN population data, much of the code is repeated.  Thus, moving the module here will remove redundant code across these country-specific repos.

In addition to this, the PR will add the following functionality:
1. The ability for users to use the time-varying forecasts from the UN WPP, up to a user-specified end date (or the maximum date available from the UN).
2. Users can enter their own custom forecasts of fertility, mortality, and immigration rates.
3. Users can enter their own estimate of the initial population distribution (Note that doing 2 and 3 together will bypass the use of UN data and use `demographics.py` purely for filling out demographic variables over the transition path and in the SS).
4. A country code parameter to specify which country's forecast to pull from the UN data.
5. Streamlines the plotting utilities and allows users to return a Matplotlib figure object (not just save a plot image file to disk).

### DIFF
--- a/docs/book/content/theory/images/SS_images.py
+++ b/docs/book/content/theory/images/SS_images.py
@@ -4,6 +4,7 @@ This script creates the plots for the steady-state chapter of the OG-Core
 documentation
 ------------------------------------------------------------------------
 """
+
 # Import libraries, packages, and modules
 import pickle
 import os
--- a/ogcore/SS.py
+++ b/ogcore/SS.py
@@ -481,9 +481,6 @@ def inner_loop(outer_loop_vars, p, client):
         debt_service,
         p,
     )
-    print("Agg tax = ", total_tax_revenue)
-    print("Agg pension outlays = ", agg_pension_outlays)
-    print("Agg UBI outlays = ", UBI_outlays)
     new_TR = fiscal.get_TR(
         Y,
         TR,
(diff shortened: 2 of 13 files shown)


jdebacker:
@rickecon All tests passing, including new ones.  According to output from `demographics.py` pop distribution seems to converge.  But I do see some odd stuff regarding immigration rates.  A few plots to illustrate:

## Fertility rates (look ok)
fert_rates_OG-Core

## Mortality rates (look ok)
mort_rates_OG-Core

## Immigration rates (too high at older ages)
imm_rates_OG-Core

## Population distribution transition (see odd uptick at 100):
PopDistPath-OG-Core

## Original vs Fixed SS population (odd uptick at 100 more apparent here):
OrigVsFixSSpop-OG-Core

I believe the uptick in the adjusted population distribution are a result of the high immigration rates at older ages.  I will review the code that backs out these rates as residuals from the transition in the actual population distribution.

jdebacker:
With the latest commits, we are looking a lot better.  Issue was that the UN  data has a bin of 100+, who were being counted as age 100, creating a mass there.  The module now drops those over age 99.

Updated plots:
OrigVsAdjImm
OrigVsFixSSpop
PopDistPath

rickecon (approved):

rickecon:
@jdebacker. This is great. Merging.

PR #897, opened 2024-01-30
Allow `chi_n` parameter to vary over the time path

jdebacker:
This PR updates OG-Core to allow users to specify a $\chi^n_s$ parameter that can vary over the time path (i.e., $\chi^n_{s,t}$). 

With this, users can simulate policies that may affect the ability of individuals to work longer at older ages.


rickecon:
@jdebacker. This looks great. Merging.

Issue #889, opened 2023-10-24
Test failures in Python 3.12.0 on Mac M1


jdebacker:
I recently rebuilt the `ogcore-dev` environment and tried to run the OG-Core testing suite. It resulted in an error when collecting tests:

```
(ogcore-dev) jason.debacker@JDEBACKER-7 OG-Core % pytest
============================= test session starts ==============================
platform darwin -- Python 3.12.0, pytest-7.4.2, pluggy-1.3.0
rootdir: /Users/jason.debacker/repos/OG-Core
configfile: pytest.ini
testpaths: ./tests
plugins: xdist-3.3.1
collecting 239 items                                                           Fatal Python error: Illegal instruction

Current thread 0x0000000200983240 (most recent call first):
  File "/Users/jason.debacker/anaconda3/envs/ogzaf-dev/lib/python3.11/site-packages/paramtools/sorted_key_list.py", line 53 in <lambda>
  File "/Users/jason.debacker/anaconda3/envs/ogcore-dev/lib/python3.12/site-packages/sortedcontainers/sortedlist.py", line 1878 in update
  File "/Users/jason.debacker/anaconda3/envs/ogcore-dev/lib/python3.12/site-packages/sortedcontainers/sortedlist.py", line 1750 in __init__
  File "/Users/jason.debacker/repos/OG-Core/ogcore/utils.py", line 393 in safe_read_pickle
  File "/Users/jason.debacker/repos/OG-Core/tests/test_output_plots.py", line 26 in <module>
  File "/Users/jason.debacker/anaconda3/envs/ogcore-dev/lib/python3.12/site-packages/_pytest/assertion/rewrite.py", line 178 in exec_module
  File "<frozen importlib._bootstrap>", line 929 in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1325 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1354 in _find_and_load
  File "<frozen importlib._bootstrap>", line 1381 in _gcd_import
  File "/Users/jason.debacker/anaconda3/envs/ogcore-dev/lib/python3.12/importlib/__init__.py", line 90 in import_module
  File "/Users/jason.debacker/anaconda3/envs/ogcore-dev/lib/python3.12/site-packages/_pytest/pathlib.py", line 567 in import_path
[...]

PR #901, opened 2024-02-12
Update Python version to be <3.12 in setup.py

rickecon:
This PR updates the Python version requirements in `setup.py` to be `python_requires=">=3.7.7, <3.12"`. The new addition is the `<3.12` piece. This requirement is in `environment.yml`. However, when OG-Core is pip installed, it uses `setup.py` and will sometimes install Python 3.12. The PR makes the Python version always be < 3.12.

cc: @jdebacker

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.11.1] - 2024-02-12 15:00:00
+
+### Added
+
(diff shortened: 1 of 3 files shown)


rickecon:
@jdebacker. This PR is ready for review. It is a small change.

jdebacker:
Thanks for looking at this @rickecon.

This seems like it relates to Issue #889.  

Because it maybe helpful going forward (i.e., making sure OG-Core is compatible with 3.12+), can you document in this PR or that issue what were the problems you noticed with the `ogcore` package in Python 3.12?

rickecon:
@jdebacker. I never verified that the errors I was getting in OG-USA were coming from Python 3.12. But the principle is that if we are restricting it in the `environment.yml` file, we need to also restrict it in the `setup.py` file.

I think we should lift that restriction or move it to `<3.13` as soon as possible. But we should only do that with testing.

Issue #900, opened 2024-02-09
size of arrays out of demographics.py


jdebacker:
Let's make the demographic parameters returned from `get_pop_objs()` that vary over the time path have their first dimension of length T + S.  Right now, some are just of length T.

jdebacker:
Resolved with PR #902

Issue #899, opened 2024-02-08
demographics.py missing an important functionality


jdebacker:
We need to allow for `demographics.py` to be able to return path of population distributions when given (1) an initial distribution (2) paths of fertility, mortality, and immigration rates.

This could probably be best implemented in the `demographics.get_pop()` function with something like:

```python
def get_pop(
    E=20,
    S=80,
    min_age=0,
    max_age=99,
    infer_pop=False
    fert_rates=None,
    mort_rates=None,
    infmort_rates=None,
    imm_rates=None,
    pre_pop_dist=None,
    country_id=UN_COUNTRY_CODE,
    start_year=START_YEAR,
    end_year=END_YEAR,
):
   if not infer_path:
       do current functionality

   else:
       find implied pop path from fert, mort, imm rates + either(pre_pop or UN data if pre pop not given)
    return pop_2D, pre_pop
```

rickecon:
This is implemented by PR #902. Closing.

PR #902, opened 2024-02-12
Extensions to `demographics.py`

jdebacker:
This PR updates the `demographics.py` module to:

1. Allow the user to have the population distribution from the initial period forward inferred from given fertility, mortality, and immigration rates (functionality to infer immigration from a given evolution of the population is retained, the user specifies what they want to do via arguments to the relevant function calls).
2. Extends all series returned from the `get_pop_objs()` function over the full transition path of T+S periods (except those that apply only to a single period).

Addresses Issues #900 and #899


rickecon (approved):

rickecon:
@jdebacker. This looks great. I ran the `run_ogcore_example.py` script on my local machine to make sure the TPI worked and converged. I will merge this as soon as the final tests pass.

You might consider updating the version to 0.11.2 in `setup.py`, `ogcore/__init__.py`, and adding a `CHANGELOG.md` entry that includes both the description at the top of the document and the GitHub compare hyperlink at the bottom of the document.

jdebacker:
@rickecon Hold off.  I want to try to figure out why the first period resource constraint isn't being satisfied and I think it's coming from `demographics.py`.  I'm still testing to find exact reason and will follow up here when I find it.

rickecon:
@jdebacker. Where is this error being manifest? Are you looking at the full TPI output after a run?

jdebacker:
@rickecon asks:
> Where is this error being manifest? Are you looking at the full TPI output after a run?

Yes.  I need set `ENFORCE_SOLUTION_CHECKS = False` in `TPI.py` so that there's not an assertion raised before the output is saved.  But when I do that and then look at the resource constraint errors along the time path, the only period with an error in excess of our tolerance is the first period.

ogcore/demographics.py
```diff
@@ -719,32 +816,91 @@ def get_pop_objs(
...
+    # assert np.allclose(pop_counter_2D, pop_dist)
+    assert np.allclose(pop_counter_2D, pop_2D)
+
+    """"
+    CHANGE - in OG-Core, we are implicitLy assuming pre-TP rates of mortality,
+    fertility, and immigration are the same as the period 0 rates.
+
+    So let's just infer the pre-pop_dist from those.
+    """
+    pop1 = pop_2D[0, :]
```
jdebacker, line 865: @rickecon This line through 882 make the fix that solves the resource constraint error I was running into from this demographics module.

In `TPI.py` we make some implicit assumptions that the immigration rate in the pre-TPI period is equal to the the rate in the first period (see [here](https://github.com/PSLmodels/OG-Core/blob/1331b6a61a2ab52ac140a59eb76a6e9a9365613f/ogcore/aggregates.py#L136)) and that the mortality rates in the pre-TPI period is equal to the mortality rates in the first period (see [here](https://github.com/PSLmodels/OG-Core/blob/1331b6a61a2ab52ac140a59eb76a6e9a9365613f/ogcore/aggregates.py#L193)).  This means that when we are using time varying mortality and fertility, we need to be very careful that everything lines up from not just the first period of TPI onward, but also from period -1 to the first model period.  

The "fix" I'm making here is one option.  It takes the assumptions embedded in `TPI.py` (that the period -1 mortality, fertility, and immigration rates are the same as in period 0) and uses that to solve for the period -1 population distribution given the population distribution in period 0.  With this, the resource constraint issue when solving TPI is resolved.

The drawback to this approach is that the period -1 population is known from the data, but we don't use it.  However, if we were to use the known period -1 population distribution and for things to all be consistent, we would need to pass the period -1 mortality, fertility, and immigration rates into the model.  That means adding three new parameters to the dictionary returned from `get_pop_objs` and in the the `Specifications` object passed around OG-Core  *or* having the same `omega`, `imm_rates`, `rho`, objects as we do now, but adding another period at the beginning (so they become length T+S+1).  However, doing that would make the indexing of those arrays different from all others in the model and therefore confusing throughout the OG-Core code (i.e., `imm_rates[0, :]` would represent immigration rates in period -1, while `Y[0]` would represent GDP in period 0.)

I opted for the approach taken here because it keeps the rest of the code base more concise and clear and the approximation error is probably not very large.  But I'm happy to chat about doing things differently.

rickecon:
@jdebacker. There is currently one test failing in `test_demographics.py::test_custom_series()` .

ogcore/demographics.py
rickecon: @jdebacker. Yes. I like this approach.

rickecon:
@jdebacker. Thanks for this PR. Looks great. Merging.

PR #904, opened 2024-02-29
[DO NOT MERGE] Log root finding

talumbau:
Put a wrapper around calls to `opt.root` so we can see how much time each call takes.


rickecon:
@talumbau @jdebacker. Closing this PR.

PR #905, opened 2024-03-02
Testing alternative SS solutions -- DO NOT MERGE

jdebacker:
Included in this PR:

1. A modification of `SS.run_SS`, which provides an argument to allow the user to specify if they would like to use the `SS.SS_fsolve` with `opt.root` to find the SS equilibrium (`fsolve=True`) or if they would like to use `SS.SS_solver` to use a bisection method to solve for the SS.
2. Adds a script to `run_examples` to test the performance of these two.

### DIFF
--- a/ogcore/SS.py
+++ b/ogcore/SS.py
@@ -1112,7 +1112,7 @@ def SS_fsolve(guesses, *args):
     return errors
 
 
-def run_SS(p, client=None):
+def run_SS(p, fsolve=True, client=None):
     """
     Solve for steady-state equilibrium of OG-Core.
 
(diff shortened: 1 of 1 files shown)


jdebacker:
My results were as follows:

```
Run time with SS fsolve =  45.24485778808594
Run time with SS Solver =  78.92088389396667
```

Also, note that the SS solver method (which does not use a root finder in the outer loop) was the only method of the two that generated performance warnings (although not many of them). e.g.:
```
2024-03-01 20:46:42,941 - distributed.utils_perf - WARNING - full garbage collections took 35% CPU time recently (threshold: 10%)
```
These results suggest that the worry I had after talking with @talumbau today -- that using parallelization through `dask` inside of a root finder was causing serious performance issues -- may not be well founded.

cc @rickecon

Issue #757, opened 2021-10-30
Fix `num_years_to_plot` in `output_plots.plot_all`


jdebacker:
Several of the plots in `output_plots.plot_all` assume that the time path is at least 150 periods (see [here](https://github.com/PSLmodels/OG-Core/blob/699a7030739e5b3f44ab4dd58fecf4cefbdc24a6/ogcore/output_plots.py#L568)).  This causes issues if a user sets `T` to a value less than 150.  

A solution to this would be to replace `num_years_to_plot=150` to `num_years_to_plot=min(base_params.T, 150)`.

sebastiansajie:
@jdebacker 
I have made the following changes to run_og_core_example.py to allow us to supply a different set of parameters to the example. I also included an option to send an email when the code has finished running.

![image](https://user-images.githubusercontent.com/6294086/140182324-0f246bc3-6030-4f2a-ad8a-6415a5e307e0.png)

![image](https://user-images.githubusercontent.com/6294086/140182535-9ae81c48-a098-4dca-b147-6e47e087c203.png)

![image](https://user-images.githubusercontent.com/6294086/140182600-6f626c09-f4a7-4b06-8e29-4ab6e33796ad.png)

This required changes in output_plots.py

![image](https://user-images.githubusercontent.com/6294086/140183485-7dc5d949-bd9f-45d5-9578-cfe4df3ffc1c.png)

The ability_bar also needed to be changed as it was accessing constants.py to access the GROUP_LABELS for the bar plot which was hard coded. I created a function to generate the labels from the lambdas.

![image](https://user-images.githubusercontent.com/6294086/140184104-d0200077-c2ff-4149-9316-ac2f3873d574.png)

jdebacker:
@sebastiansajie I like your fixes to the `output_plots.py.plot_all()` function and the `lambda_labels` (really creative solution there!).  Thank you for proposing these.

I'd recommend adding the `lambda_labels()` function to `utils.py` as it it can then be imported and used in `output_plots.py` and `parameter_plots.py` (both of which currently import the `GROUP_LABELS` from `constants.py` and therefore have issues without `J` in [7, 9, 10].

I think the changes your propose to the example script are best left out of OG-Core for now.  I like having the example script produce output with a more realistic calibration, even if it does take more time to run.  You can just keep your modified script around for your own use (I have a bunch of these for different simulations I've run).

[...]

PR #923, opened 2024-04-14
Generalizations in `output_plots.py`

jdebacker:
This PR does to things to make `output_plots.py` work better with different parameterizations:

1. It ensures that the `num_years_to_plot` is less than `T` (or raises an assertion if not).
2. Labels for ability types are generalized to work with any `lambdas` values.

Addresses Issue #757


rickecon (approved):

rickecon:
This looks good to me. Thanks @jdebacker. Merging.

PR #957, opened 2024-08-03
Fix bug in the `baseline_spending=True` parameterization

jdebacker:
In PR #880, all TPI output arrays were cast as have a length of `T` in the time dimension.

This posed an error for the case of `baseline_spending=True`, which was looking for aggregate transfers (`TR`) to be of length `T+S`.  This PR fixes that error.


rickecon (approved):

rickecon:
@jdebacker. This looks great. Now passed all tests. Merging now.

PR #969, opened 2024-08-16
Allow for compatibility with Python 3.12

jdebacker:
This PR updates test files so that the unit tests pass on Python 3.12.  As noted in Issue #889, the OG-Core has run on Python 3.12 for some time.  The update here ensure that we can tests against 3.12 and therefore publish packages compatible with that version of Python.

cc @rickecon

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.10", "3.11", "3.12"]
+        python-version: ["3.11", "3.12"]
 
     steps:
       - name: Checkout
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.10", "3.11"]
+        python-version: ["3.10", "3.11", "3.12"]
 
     steps:
       - name: Checkout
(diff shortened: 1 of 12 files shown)


jdebacker:
Remaining work:

* I am aware of a test failure in the `test_output_tables.py` function.  It seems that the new 3.12 pickle from has just one industry whereas the TPI output read into the `test_output_tables.test_dynamic_revenue_decomposition` test has three industries.  TODO: update either the 3.12 pickle to have `M=3`.

rickecon:
@jdebacker. The black formatting test is failing. And we should update the middle number (minor) of the version for this one (e.g., 0.12.0). And I think we should make the version requirement `python>= 3.7.7, <3.13` in both `environment.yml` and `setup.py`. And finally, if we move to Python 3.12 as our latest supported version, I think we should change our GitHub Actions to run on Python 3.11 and 3.12, with the single runs running on 3.12.

rickecon:
@jdebacker.
- The four tests at the end of `test_output_tables.py` are failing in MacOS Python 3.12, but not in Python 3.10 and 3.11. I thought this was fixed with your previous commits. So I don't understand why it is failing now.
- In `build_and_test.yml`, I think we should only test two versions of Python. So I propose we update the tests to only by Python 3.11 and 3.12.
- In `environment.yml` and `setup.py`, I think we should limit the Python version to be less than Python 3.13 (python<=3.7.7,<3.13`). This will prevent the problems that inevitably arise when someone's conda package allows Python 3.13. [This page](https://devguide.python.org/versions/) shows the production schedule and expiration dates of Python packages. I think this highlights the value of just keeping two packages that we test on. Note that Python 3.10 is set to expire on 2026-10.
- In `check_black.yml`, `deploy_docs.yml`, `docs_check.yml`, and `publish_to_pypi.yml`, update the Python version line to `python-version: "3.12"`.
- In `setup.py`, `__init__.py`, and `CHANGELOG.md`, update the version to `0.12.0`.

rickecon (approved):

rickecon:
@jdebacker. This looks great. Thanks for making this update. Merging now.

PR #990, opened 2024-09-16
[WIP] HH solution first draft

john-p-ryan:
This adds a first draft of the household problem solution with stochastic income shocks using EGM. 

A few notes about the solution method and modeling choices:

1. At the end of life, the household solution boils down to 1 variable in 1 unknown to be solved numerically if solved forwards. That is, for each point in the capital and productivity grid, we make a guess of labor supply (which is known to be in the compact interval $\[0, \tilde{l} \]$). We then infer consumption from labor using the labor supply Euler equation. Lastly, the bequests left at the end of life are solved using the end of life savings Euler equation. This implies a residual in the budget constraint, for which a root is solved numerically.
2. The rest of the life cycle, the household solution is a system of 2 equations in 2 unknowns to be solved numerically and using the endogenous grid method. This takes savings one period ahead and productivity as given, and for each point in the capital and productivity grid, we solve for current period consumption using the savings Euler equation and policy functions in the next period. Then, we make a guess of current period assets and labor supply, which implies a residual in the labor supply FOC and budget constraint, which is solved numerically. Using the solution, we interpolate across the asset grid to recover current period policy functions.
3. Currently, this solution keeps the Elliptical disutility of labor supply and CRRA bequest motive as they maintain their computational advantage since you don't have to check for binding constraints and resolve when some constraints are binding (borrowing constraint without CRRA bequest, labor supply endowment with Frisch disutility). This can be relaxed later. 

There is still much to do. Current priorities as follows:
1. Finish writing tests for each of the new functions
2. Ensure tax functions are used correctly
3. Ensure dimensions are correct when on transition path vs steady state

After these are finished, we can move on to incorporating this solution into `aggregates.py`, `SS.py` and `TPI.py`.

### DIFF
--- a/ogcore/household.py
+++ b/ogcore/household.py
@@ -6,6 +6,7 @@ Household functions.
 
 # Packages
 import numpy as np
+import scipy.optimize as opt
 from ogcore import tax, utils
 
 """
@@ -14,23 +15,6 @@ from ogcore import tax, utils
 ------------------------------------------------------------------------
 """
 
-def mu_c(c, sigma):
-    r"""
-    Compute the marginal utility of consumption.
-
-    .. math::
-        MU_{c} = c^{-\sigma}
-
-    Args:
-        c (array_like): household consumption
-        sigma (scalar): coefficient of relative risk aversion
-
-    Returns:
-        output (array_like): marginal utility of consumption
-
-    """
-
-    return c**(-sigma)
 
 def marg_ut_cons(c, sigma):
     r"""
@@ -60,26 +44,6 @@ def marg_ut_cons(c, sigma):
     output = np.squeeze(output)
 
     return output
-
-
-def mu_n(n, chi_n, p):
-    r"""
-    Compute the marginal disutility of labor.
-
-    .. math::
-        MDU_{n} = \chi^n_{s} n^{\frac{1}{\theta}}
-
-    Args:
-        n (array_like): household labor supply
-        chi_n (array_like): utility weights on disutility of labor
-        p (OG-Core Specifications object): model parameters
-
-    Returns:
-        output (array_like): marginal disutility of labor supply
-
-    """
-    return chi_n * n ** (1 / p.theta)
-
 
 def marg_ut_labor(n, chi_n, p):
     r"""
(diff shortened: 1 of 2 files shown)


ogcore/household.py
```diff
@@ -769,3 +816,274 @@ def constraint_checker_TPI(b_dist, n_dist, c_dist, t, ltilde):
...
+        t (int): model period
+        j (int): index of ability type
+        p (OG-Core Specifications object): model parameters
+        method (str): adjusts calculation dimensions based on 'SS' or 'TPI'
+        
+        Returns:
+            BC_error (array_like): residuals of the household budget constraint"""
+    c = c_from_n(n, b, p_tilde, r, w, factor, e, z, chi_n, etr_params, mtrx_params, t, j, p, method)
+    b_splus1 = b_from_c_EOL(c, p_tilde, j, p.sigma, p)
+    net_tax = tax.net_taxes(r, w, b, n, bq, factor, tr, ubi, theta, t, j, False, method, e, etr_params, p)
```
jdebacker, line 894: Should it be `e*z` in the arguments to the `tax.net_taxes()` call?

ogcore/household.py
```diff
@@ -769,3 +816,274 @@ def constraint_checker_TPI(b_dist, n_dist, c_dist, t, ltilde):
...
+        j (int): index of ability type
+        t (int): model period
+        p (OG-Core Specifications object): model parameters
+        method (str): adjusts calculation dimensions based on 'SS' or 'TPI'
+        
+        Returns:
+            HH_error (array_like): residuals of the household budget constraint and labor supply Euler equation"""
+    b = x[0]
+    n = x[1]
+    net_tax = tax.net_taxes(r, w, b, n, bq, factor, tr, ubi, theta, t, j, False, method, e, etr_params, p)
```
jdebacker, line 946: Should it be `e*z` in the arguments to the `tax.net_taxes()` call?

vahid-ahmadi:
WIP, so just one structural note. The EGM formulation reads well and keeping elliptical disutility plus the CRRA bequest to avoid constraint checking is a sensible place to start.

The thing worth managing now is the rename of household.py's public API — `marg_ut_cons` to `mu_c`, `FOC_savings` to `c_from_b_splus1`, the new `z` argument on `get_cons`. The branch is clean against `stochastic_income`, but that base hasn't moved since September 2024 while household.py on master has, so the collision is deferred rather than avoided. Merging master in periodically would keep that cheap.

The per-gridpoint `opt.root` loop in `solve_HH` will need a look on performance grounds, but that can reasonably wait for SS/TPI integration.

PR #1037, opened 2025-06-25
Make interest rate on government debt a function of the debt to GDP ratio

jdebacker:
Since the risk premium of governments may vary with the amount of debt they accumulate, this PR updates the `fiscal.get_r_gov` function to allow it to vary with the debt to GDP ratio in a quadratic manner.  The new specification adds to parameters to the function determining `r_gov`. The new function is:
$r_{gov,t}= (1-\tau_{d,t})r_{t} - \mu_d + \beta_1 \frac{D_t}{Y_t} + \beta_2 \left(\frac{D_t}{Y_t}\right)^2$

In the code $\beta_1$ maps to `r_gov_DY` and $\beta_2$ maps to `r_gov_DY2` to avoid confusion with the rate of time preference.

### DIFF
--- a/tests/test_fiscal.py
+++ b/tests/test_fiscal.py
@@ -11,44 +11,44 @@ CUR_PATH = os.path.abspath(os.path.dirname(__file__))
 df = pd.read_csv(
     os.path.join(CUR_PATH, "test_io_data", "get_D_G_path_data.csv")
 )
-Y = df["Y"].values
-TR = df["TR"].values
-Revenue = df["Revenue"].values
-Gbaseline = df["Gbaseline"].values
-D1 = df["D1"].values
-D2 = df["D2"].values
-D3 = df["D3"].values
-D4 = df["D4"].values
-G1 = df["G1"].values
-G2 = df["G2"].values
-G3 = df["G3"].values
-G4 = df["G4"].values
-D_d1 = df["D_d1"].values
-D_d2 = df["D_d2"].values
-D_d3 = df["D_d3"].values
-D_d4 = df["D_d4"].values
-D_f1 = df["D_f1"].values
-D_f2 = df["D_f2"].values
-D_f3 = df["D_f3"].values
-D_f4 = df["D_f4"].values
+Y = df["Y"].values.copy()
+TR = df["TR"].values.copy()
+Revenue = df["Revenue"].values.copy()
+Gbaseline = df["Gbaseline"].values.copy()
+D1 = df["D1"].values.copy()
+D2 = df["D2"].values.copy()
+D3 = df["D3"].values.copy()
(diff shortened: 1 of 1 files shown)


rickecon:
@jdebacker. Thanks for submitting this PR. Great update. Here is one suggestion and two questions.
- This is a relatively small addition to OG-Core. I recommend that this be a patch version update to 0.14.5, rather than a minor version update to 0.15.0 (see [Semantic Versioning](https://semver.org/spec/v2.0.0.html)).
- In your `fiscal.py`, the `get_r_gov()` function takes `DY_ratio` as an argument. Why in your `TPI.py` in line 707 does `r_gov` take 0 as the `DY_ratio` value?
- Have you run an example locally with `r_gov_DY>0` and `r_gov_DY2>0`?

ogcore/TPI.py
```diff
@@ -704,7 +704,7 @@ def run_TPI(p, client=None):
     total_tax_revenue = np.ones(p.T + p.S) * ss_vars["total_tax_revenue"]
 
     # Compute other interest rates
-    r_gov = fiscal.get_r_gov(r, p, "TPI")
+    r_gov = fiscal.get_r_gov(r, 0, p, "TPI")
```
rickecon, line 707: @jdebacker. Why is the `DY_ratio` argument in this line equal to 0? You have the `D` and the `Y` series above.

jdebacker: I've updated this to take the initial guess at the debt to GDP path (although since these are initial guesses, they shouldn't affect the model solution, but make affect the time to convergence).

jdebacker:
I've updated this PR to a WIP as I'm not seeing all my local tests pass.  In particular, there test failures for the `run_TPI` function.  In `test_run_TPI_full_runs` the following parameterizations are failing:

* `reform`
* `Baseline, delta_tau = 0`
* `Baseline, Kg > 0` (even though the `Baseline, M=3 non-zero Kg` parameterization passes)

In each of these, I notice that all the aggregate values and prices are within the 1e-4 test tolerance *except* the household variables (`c`, `n`, `b_sp1`, etc.).  Difference seem larger around they year of the closure rule, but there are some other large values in other years.  All the largest differences appear to come from the lowest and highest ability types, with more than 95% of the differences above the threshold from the highest ability types.

I am puzzled why this happens for these parameterizations only -- and why this happens as at since other tests (e.g., in `test_fiscal.py`) show that these changes have no effect on the path of `D`, `G`, `r_gov` or other variables from the changes to `fiscal.py`.

cc @rickecon

rickecon:
@jdebacker. This looks great. Thanks for adding those `.copy()` commands in `test_fiscal.py`. Merging.

Issue #1039, opened 2025-07-08
Get rid of initial_guess_w_SS parameter


rickecon:
We should get rid of the `initial_guess_w_SS` parameter and specify our solution method correctly in the steady-state equilibrium documentation of OG-Core.
    - In our steady-state computation in `SS.py`, we no longer use the `initial_guess_w_SS` parameter.
    - Instead, we solve for our initial guess for `w_SS` as a function of our initial guess for `r_SS` in [line 1289 of SS.py](https://github.com/PSLmodels/OG-Core/blob/master/ogcore/SS.py#L1289) using the `get_w_from_r()` function in `firm.py`.
    - The steady-state equilibrium solution documentation in the "[Steady-state solution method](https://pslmodels.github.io/OG-Core/content/theory/equilibrium.html#steady-state-solution-method)" section of the OG-Core documentation does not specify this part of generating the initial guess for w from the initial guess for r. Nor does it specify that the initial guess for r_p comes from the initial guess for r.

Note: If we remove the `initial_guess_w_SS` parameter from OG-Core, we will have to remove it from any of our country calibrations that have it in their respective `[cntry_package]_default_parameters.json` file. For example, this parameter exists in the OG-PHL file currently.

cc: @jdebacker

jdebacker:
Resolved in PR #1042.

PR #1042, opened 2025-08-03
Remove `initial_guess_w_SS` from default params

jdebacker:
This PR removes `initial_guess_w_SS` from default params per suggestions in Issue #1039.


.github/workflows/build_and_test.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.11", "3.12"]
+        python-version: ["3.12", "3.13"]
```
rickecon, line 29: I recommend that we do not update to supporting Python 3.13 and 3.12 until Mar.-Apr. 2026 when Python 3.13 is scheduled to come out of "bugfix" mode (see "[Status of Python Versions](https://devguide.python.org/versions/)" page). I made a proposal today of a "Python version policy" for OG-Core and its country calibration repositories in [this OG-Core discussion](https://github.com/PSLmodels/OG-Core/discussions/1046). I would love to hear your thoughts on this.

rickecon: OK. Let's move to Python 3.12 and 3.13. @jdebacker

rickecon (approved):

PR #1073, opened 2025-11-16
Better handle the pre-time path population distribution

jdebacker:
The population distribution before the start year of the model is necessary to compute savings and bequests in period 0.

This PR adds explicitly the parameters for mortality, immigration, and population growth rates from period 0 to period 1 in order to make for a consistent transition from this "pre model period" to period 0.

cc @rickecon


SeaCelo:
@jdebacker @rickecon: I dug into the period-1+ drift you flagged. The boundary fix idea is right — master does have a real period-0 inconsistency — but the drift across the rest of the path is coming from how growth rates are indexed in this branch.

In master, `g_n[t]` is the growth from `t-1` to `t`. In this branch it's the growth from `t` to `t+1`. The aggregate formulas in `aggregates.py` weren't updated for the new convention, so they read the wrong period's growth rate everywhere.

You can confirm this without running the model. The population accounting itself fails the consistency check across t=1 to fixper, in the exact shape you reported.

The reindex is a one-line revert. But this PR also bundles other changes that aren't strictly part of the boundary fix — the timeline extension, the slicing on returned arrays, the new `fixper` formula, the `pre_pop_dist` removal. Rather than untangle them, I'd rather propose a smaller PR with just the boundary fix. I'll open it shortly and link back.

SeaCelo:
Opened #1117 with the smaller version.

marcelolafleur:
@jdebacker — here's what I think is going on and why you're seeing that drift.

This PR changes one line in `get_pop_objs` that stores the population growth rates. Master writes them starting at position 1 of the array; this PR writes them starting at position 0. That moves every growth rate one slot earlier in the array, and the boundary value — the growth rate from the year before the model starts to the first model year — that used to sit at the front of the array now lives in the new `g_n_preTP` field instead.

The drift is caused by `aggregates.py` not being fully updated to match. The steady-state code was updated to read from `g_n_preTP` (lines 137 and 192). The transition-path code wasn't. It still pulls the growth rate from the same positions in the array as master, but those positions now hold a different year's value, and the new `g_n_preTP` field is never read on the transition-path side. So every transition-path year picks up a growth rate that's one year off, and the boundary value never reaches the formula at all.

The transition-path updates would be:

1. `aggregates.py` line 160 (`get_B` TPI). The growth-rate slice currently starts at position 1, which was the right place under master's storage. Under this PR's storage, the same slice now starts a year too late, so shift it back to position 0:
   - Replace `np.hstack((p.g_n[1 : p.T], p.g_n_ss))` with `np.hstack((p.g_n[: p.T - 1], p.g_n_ss))`.

2. `aggregates.py` lines 215 and 222 (`get_BQ` TPI). The growth-rate slice currently starts at position 0 to pick up the boundary value. Under this PR's storage, the boundary value isn't in the array anymore — it's in `g_n_preTP`. So prepend `g_n_preTP` and shift the rest of the slice back by one:
   - Replace `1.0 + p.g_n[: p.T]` with `1.0 + np.append(p.g_n_preTP, p.g_n[: p.T - 1])`.

`get_I` reads the growth rate the same way at lines 98, 100, and 105 — those would need the same shift to match the new storage layout.

(The test expected values in `test_aggregates.py` also encode master's indexing, so they'd need updating to use `g_n_preTP` at the boundary too.)

cc @rickecon

marcelolafleur:
Following up on our offline discussion. Ran the boundary fix end-to-end in all 5 country repos — full TPI baselines, each with its own `<og>_default_parameters.json` + real UN demographics through `update_specifications`.

Three branches tested per country: `upstream/master` (current code), this PR rebased onto current `upstream/master`, and the fix (PR #1117 or this PR with the May 11 `aggregates.py` edits applied — same math).

| Country | Master t=0 (bug) | PR #1073 t=0 / max drift t≥1 | Fix t=0 (clean) |
|---|---:|---:|---:|
| OG-USA | -3.46e-03 | 4.5e-05 / **2.0e-03** | +3.62e-08 |
| OG-ETH | -1.22e-03 | 2.7e-05 / **3.1e-03** | +8.87e-07 |
| OG-ZAF | -3.39e-03 | 1.2e-04 / **6.6e-03** | -5.58e-09 |
| OG-PHL | +2.05e-03 | 9.4e-05 / **1.9e-02** | -1.49e-06 |
| OG-IDN | +7.04e-04 | 9.1e-06 / **2.6e-03** | -2.54e-09 |

Master (red) has the t=0 boundary inconsistency and is clean at t≥1. PR #1073 as-is (orange) fixes t=0 but introduces sustained drift at t≥1 — trading one error for another, which is the issue you reported on OG-USA in December. With the `aggregates.py` edits from my May 11 comment (blue), t=0 is fixed *and* the drift is avoided.

<table>
<tr>
<td></td>
<td></td>
</tr>
<tr>
<td></td>
<td></td>
</tr>
<tr>
<td colspan="2" align="center"></td>
</tr>
</table>

**How to apply the fix in this PR.** Six line edits in `aggregates.py`:

- Lines 98, 100, 160: `np.hstack((p.g_n[1 : p.T], p.g_n_ss))` → `np.hstack((p.g_n[: p.T - 1], p.g_n_ss))`
- Lines 215, 222: `1.0 + p.g_n[: p.T]` → `1.0 + np.append(p.g_n_preTP, p.g_n[: p.T - 1])`
- Line 105: `p.g_n[1 : p.T + 1]` → `p.g_n[: p.T]`

Plus the expected-value calculations in `test_aggregates.py` lines 183 and 239 need the same indexing update — they currently use master's pattern and would mismatch the corrected `get_B`/`get_BQ` output, breaking the existing tests.

**These changes are all in OG-Core — no per-country PRs needed.** Every downstream country repo runs `Calibration → get_dict → update_specifications`, so once `get_pop_objs` returns the new preTP fields and `aggregates.py` reads them correctly, the boundary fix takes effect automatically across all of them.

**Test setup per country.** Loaded each country's `<og>_default_parameters.json`, then called `ogcore.demographics.get_pop_objs(country_id=<UN_M49>)` directly to pull real year-(-1) demographics. Same outcome as `Calibration(update_from_api=True)` but skips the macro/income APIs that aren't relevant to the boundary test (the PHL World Bank/IMF path returned values that broke SS on master; isolating to just the demographic path sidestepped that). For OG-ETH I used `country_id="231"` — `ogeth/calibrate.py` hardcodes `"710"` (ZAF M49), a separate bug worth flagging.

This PR (#1073) is the cleaner shape — pulling the boundary value out at the source rather than layering it through `update_specifications` like #1117 does.

I don't have push access to this branch, so #1117 is my fallback with the same math. I'll post the same results there.

cc @rickecon @jdebacker

jdebacker:
@SeaCelo Thank you for all your work on this.  I think you've put this on the right track.  Current tests with this branch as satisfactory to me:

```
Max RC error USA:  0.0003460672290931596
Max RC error ETH:  0.00040633855858390744
Period 0-5 RC error USA:  2.436032920244413e-07
Period 0-5 RC error ETH:  0.00040633855858390744
(base) jason.debacker@JDEBACKER-8 OG-Core % uv run python check_prepop_results.py
Max RC error USA:  0.0003460672290931596
Max RC error ETH:  0.00040633855858390744
Period 0-5 RC error USA:  [[ 2.43603292e-07]
 [ 1.87356859e-07]
 [ 1.81543269e-07]
 [ 1.33740551e-07]
 [ 6.91632409e-08]
 [-4.36452955e-09]]
Period 0-5 RC error ETH:  [[ 1.95733206e-06]
 [ 1.43801852e-06]
 [-4.06338559e-04]
 [-1.29647649e-07]
 [-3.76461040e-07]
 [-5.14581436e-07]]
Period -1 RC error USA:  [-3.74973919e-07]
Period -1 RC error ETH:  [-1.43628822e-06]
```

marcelolafleur:
One thing to flag before this merges, since it touches the country repos: a couple of the demographics functions changed shape.

- `get_pop` now returns just the population path — it used to also return the prior-year distribution (two values, now one).
- `get_pop` and `get_pop_objs` no longer take the `pre_pop_dist` argument.

The normal calibration path isn't affected — it reads the new pre-period fields automatically through `update_specifications`. It's only code that calls these two functions directly, or notebooks still passing `pre_pop_dist`, that'll need a quick update. Might be worth a CHANGELOG note and bumping the country repos in step so the upgrade stays clean.

jdebacker:
@SeaCelo Thanks for these comments.  I've added the note about API changes to the change log, as you suggest.  Once this PR is merged, I will go through the country repos and make sure use of the demographics module is consistent with the new `demographics.py`.

marcelolafleur:
@jdebacker this is a great fix. Heads up that I made a small PR to this, fixing some display issues:  https://github.com/jdebacker/OG-Core/pull/42

jdebacker:
Local tests all passing:
```
tests/test_SS.py ..........................................              [  6%]
tests/test_TPI.py .....................................                  [ 12%]
tests/test_aggregates.py ..........................................      [ 19%]

tests/test_basic.py ....                                                 [100%]
tests/test_demographics.py ..................                            [100%]
tests/test_elliptical_u_est.py .......                                   [100%]
tests/test_execute.py .                                                  [100%]
tests/test_firm.py ..................................................... [ 76%]
................                                                         [100%]
tests/test_fiscal.py ..........................                          [100%]
tests/test_household.py ................................................ [ 94%]
...                                                                      [100%]
tests/test_output_plots.py ............................................. [ 90%]
.....                                                                    [100%]
tests/test_output_tables.py ...................                          [100%]
tests/test_parameter_plots.py .......................................... [ 97%]
.                                                                        [100%]
tests/test_parameter_tables.py .......                                   [100%]
tests/test_parameters.py ...............                                 [100%]
tests/test_pensions.py .................................                 [100%]
tests/test_population_accounting.py ..                                   [100%]
tests/test_run_example.py ..                                             [100%]
tests/test_run_ogcore.py .                                               [100%]
tests/test_tax.py ...................................                    [100%]
tests/test_txfunc.py ...................................                 [100%]
tests/test_user_inputs.py .........                                      [100%]
tests/test_utils.py .................................................... [ 62%]
...............................                                          [100%]
```

PR #1074, opened 2025-11-18
Fix consumption tax revenue calculation

jdebacker:
This PR fixes the consumption tax revenue under differentiated goods specifications by passing the `c_i` object rather than the `c_mat` object in SS and TPI calculations.


rickecon:
@jdebacker. This looks great. I just fixed one typo in the `CHANGELOG.md`. I'll merge as soon as tests pass.

PR #1077, opened 2025-12-03
Use baseline theta

jdebacker:
This PR updates OG-Core with a new parameter, `baseline_theta`.  If `baseline_theta=True`, then the reform run uses the same replacement rate in the Social Security pension system as was computed in the baseline simulation.


rickecon:
@jdebacker. I pulled this branch, updated the conda environment, and ran the `run_ogcore_example.py` script. Script ran smoothly through all phases of baseline and reform, steady state and time path. I am merging this.

Issue #1080, opened 2025-12-11
Reduce memory of the parameters class object


jdebacker:
For a typical calibration, the parameters class object is not huge (~20MB), but if one were doing large scale parallelization (or running on a machine with low memory), this could cause issues.

I think we can reduce the footprint of this object without loosing much.  A couple of ideas:
1. Use the more efficient data types (e.g. `retirement_age` can be cast as an integer, more floating point values would be fine with 32 bit (or even 16 bit).
2. Most parameters become constant long before the steady state: is there a way to reduce the size of arrays knowing that after some point in the `T` dimension, things stop changing?

jdebacker:
Related: [Example of custom data type in ParamTools](https://github.com/PSLmodels/ParamTools/blob/3a9094f50b45b5ec7324d8ffa9d6ef7070957452/docs/api/custom-types.ipynb#L14)

PR #1083, opened 2025-12-22
Downcast parameter data types

jdebacker:
This PR attempts to address Issue #1080, but using more memory-efficient datatypes in the `Specifications` class object: 16-bit integers and 32-bit floating point numbers replace 64-bit integers and 64-bit floating point numbers.


vahid-ahmadi:
The custom paramtools field is the right hook for this, and keeping solver tolerances at float64 was the correct call.

Two things before it goes further. It's currently conflicting with master, and master has added parameters since December, so the conversion now covers 89 of ~101 floats — worth a rebase plus a stated rule for which type new params get, otherwise this drifts again. More substantively: float32 params upcast to float64 the moment they touch model arrays, so the saving is confined to the `Specifications` object itself while expected-value tests shift at ~1e-7. Probably worth measuring the actual memory delta against that test churn before committing to it.

Minor: the description says 16-bit integers, the code registers int32.

PR #1091, opened 2026-02-25
fix: replace bare except with specific exceptions in solve_L, fix adj_cost docstring typo

suyash469:
## Summary
Two small real bug fixes in [ogcore/firm.py](cci:7://file:///e:/gsoc-ogclews/OG-Core/ogcore/firm.py:0:0-0:0).

### 1. [solve_L()](cci:1://file:///e:/gsoc-ogclews/OG-Core/ogcore/firm.py:654:0-708:12) — Replace bare `except:` with `except (ValueError, TypeError):`
The `try/except` block exists solely to detect whether `K_g` is a NumPy array
(comparing an array to a scalar raises `ValueError`) or an incompatible type
(`TypeError`). A bare `except:` silently catches `KeyboardInterrupt`, `SystemExit`,
and `MemoryError`, making long OG-Core runs impossible to interrupt and masking
real errors silently.

### 2. [adj_cost()](cci:1://file:///e:/gsoc-ogclews/OG-Core/ogcore/firm.py:711:0-735:14) — Fix docstring typo
`'adjstment'` → `'adjustment'`

## Validation
- Confirmed zero bare `except` clauses remain in [firm.py](cci:7://file:///e:/gsoc-ogclews/OG-Core/ogcore/firm.py:0:0-0:0) via AST check
- Typo verified absent from source

### DIFF
--- a/ogcore/SS.py
+++ b/ogcore/SS.py
@@ -631,7 +631,7 @@ def inner_loop(outer_loop_vars, p, client):
     )
 
     G_vec = np.zeros(p.M)
-    G_vec[-1] = G
+    G_vec[-1] = float(G)
     C_m_vec = np.dot(p.io_matrix.T, C_vec)
     I_d_vec = np.zeros(p.M)
     I_d = aggr.get_I(b_splus1, K_d, K_d, p, "SS")
--- a/ogcore/SS.py
+++ b/ogcore/SS.py
@@ -529,7 +529,7 @@ def inner_loop(outer_loop_vars, p, client):
     theta = pensions.replacement_rate_vals(nssmat, new_w, new_factor, None, p)
 
     # Find updated goods prices
-    new_p_m = firm.get_pm(new_w, Y_vec, L_vec, p, "SS")
+    new_p_m = firm.get_pm(new_w, Y_vec, L_vec, p, "SS", vectorized=True)
     new_p_m = new_p_m / new_p_m[-1]  # normalize prices by industry M
     new_p_i = np.dot(p.io_matrix, new_p_m)
     new_p_tilde = aggr.get_ptilde(new_p_i, p.tau_c[-1, :], p.alpha_c)
--- a/ogcore/SS.py
+++ b/ogcore/SS.py
@@ -298,7 +298,7 @@ def inner_loop(outer_loop_vars, p, client):
                 schema_backup[attr] = getattr(p, attr)
                 try:
                     delattr(p, attr)
-                except:
+                except AttributeError:
                     pass
 
         # Scatter the parameters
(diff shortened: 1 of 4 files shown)


jdebacker (approved):

arihantlodha-cmd (commented):
Nice PR. The exception fixes are the right call: catching bare `except:` was genuinely hiding `KeyboardInterrupt`/`SystemExit` on long runs, and the specific exceptions match the intent (`AttributeError` for the delattr/setattr, `(ValueError, TypeError)` for the K_g array check in `solve_L`). I also like the `np.any(x == 0)` fix in `get_MPx`: the old `np.any(x) == 0` was actually wrong (it asked whether "any element is truthy" equals 0, i.e. whether all elements are zero), so that's a real correctness fix.

I'm newer here so treat this as non-blocking, but a few things stood out.

**The title undersells the PR.** It reads as two small bug fixes, but the substantive change is vectorizing `get_MPx` and `get_pm` across industries and switching all the `get_pm` call sites to `vectorized=True`. That's a real change to a production-function hot path, not a cleanup, and someone reviewing on the title alone might not expect the pricing math to move. Might be worth splitting the exception fixes into their own tiny PR (fast to merge) and letting the vectorization stand on its own, or at least re-describing so the vectorization is front and center.

**`get_pm`'s SS return type changed.** In the non-vectorized SS branch it now returns `float(pmout[0])` instead of an array. The in-repo callers are all switched to `vectorized=True` so they're fine, but a downstream caller (a country model) doing `get_pm(w, Y, L, p, "SS")` would silently get a scalar instead of the array it used to get. Worth a changelog note at least, since it's a quiet interface change.

**Is there a test that the vectorized path equals the old loop?** The existing SS/TPI regression tests should catch a divergence if they run, but an explicit `np.allclose(get_pm(..., vectorized=True), <per-industry loop>)` would make the refactor easy to trust and guard it going forward.

Minor: `get_MPx`'s default `m` changed from -1 to 0, which is fine since every caller passes `m` explicitly, but the docstring still says "when `m=-1` the function computes it for the last industry," so the doc and the default disagree now.

Thanks for this, the interruptibility fix alone is worth it.

rickecon:
@suyash469. Sorry to be slow to get to this PR. And thanks for taking the time to start it. The bare `except` in `solve_L()` was updated in PR #1093. But the typo in the `firm.adj_cost()` function description still needs to be fixed. And I am interested in the vectorization discussion, but I think that functionality already exists in the model functions. As noted by @arihantlodha-cmd, we do need to update the documentation. And we do need to update some of the docstrings in `firm.py`. I am happy to revisit this PR if you have the time and availability. Otherwise, we will incorporate these changes in a new PR. I am closing this PR for now due to it requiring many changes and it being old. But I am very open to reopening it and spending the time to get it through. We really appreciate your participation and effort and would love for you to have a merged PR.

cc: @jdebacker

PR #1096, opened 2026-03-31
Add numerical guards for CES with heterogeneous epsilon

vahid-ahmadi:
## Summary
- Adds `_FLOOR = 1e-12` guards in `firm.py` to prevent `0^(negative)` → inf/NaN when CES epsilon ≠ 1
- Fixes bug in `get_MPx`: `np.any(x) == 0` should be `np.any(x == 0)` (the original always evaluates to False)
- Guards applied in `get_Y`, `get_MPx`, `get_KLratio_KLonly`, and `solve_L` — only in the general CES code paths; Cobb-Douglas (ε=1) is unchanged

## Motivation
When `epsilon` varies across sectors (e.g., Energy=0.50, Real Estate=0.40, Info & Finance=1.20), the CES production function raises K, L, K_g to `(epsilon-1)/epsilon`. For epsilon < 1, this exponent is negative, so any zero or near-zero input during initial solver iterations produces inf/NaN and crashes the solver.

This was blocking OG-UK from using calibrated sector-specific CES elasticities from Chirinko (2008) and Knoblach et al. (2020). With this fix, OG-UK's 8-sector model with heterogeneous epsilon (0.40–1.30) converges successfully in both SS and TPI.

## Test plan
- [ ] Existing OG-Core tests pass (guards only activate when inputs ≈ 0)
- [ ] OG-UK 8-sector SS with heterogeneous epsilon converges (tested, resource constraint ~1e-6)
- [ ] Single-sector (M=1, ε=1.0) results unchanged (guards not activated for Cobb-Douglas)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jdebacker (approved):

vahid-ahmadi:
Fixed the failing tests (`test_get_KLratio_from_r[epsilon=0.5,TP]` and `test_get_KLratio[epsilon=0.5,TP]`).

**Root cause:** `np.maximum(denom, _FLOOR)` in `get_KLratio_KLonly` was clamping legitimately negative denominators to a small positive value. For `epsilon=0.5` with certain interest rates, `denom` goes negative (valid — indicates an economically infeasible region), but the floor was silently flipping it positive, producing near-zero KLratios instead of the correct values.

**Fix:** Replaced with a sign-preserving guard:
```python
denom = np.where(
    denom >= 0,
    np.maximum(denom, _FLOOR),
    np.minimum(denom, -_FLOOR),
)
```
This still prevents division by zero but preserves negative denominators.

jdebacker (approved):

jdebacker:
Thanks for the PR @vahid-ahmadi.  Merging.

PR #1122, opened 2026-05-08
Update uv.lock and version

rickecon:
This PR:
- Updates the `uv.lock` to include dependabot upgrades from PR #1118, PR #1119, PR #1120, and PR #1121, as well as some other package upgrades that were available.
- Upgrades the version in `pyproject.toml` and `__init__.py`.

cc: @jdebacker


jdebacker (approved):

jdebacker:
@rickecon Thanks for updating the `uv.lock` file.  I'll merge this, but we should have a discussion about whether updates like this trigger a new version. Since the `uv.lock` file is not part of the `ogcore` package, I don't think a new release is necessary.

rickecon:
@jdebacker. I really just wanted a fully updated version for the Ethiopia training. Thx.

PR #1128, opened 2026-05-14
3.3x speedup: cache e_long, skip redundant tax-parameter conversions

SeaCelo:
`FOC_savings` and `FOC_labor` rebuild an array called `e_long` on every call — an `np.concatenate` of `p.e` with a tiled copy of its last row. Profiling a single-reform TPI run with py-spy across all workers showed this one construction was the single most expensive operation in the run.

`e_long` is a pure function of `p.e`, `p.S`, and `p.J`. None of those change during a solve, so there is no reason to rebuild it. This PR adds a `_get_e_long(p)` helper that builds the array once and caches it on the parameters object; `FOC_savings` and `FOC_labor` call the helper instead of constructing the array inline. There are four call sites in total — the `j is not None` and `j is None` branches of each function — and all four now use the helper.

Because `p` is scattered once per TPI solve and Dask hands each worker the same materialized object across tasks, `e_long` is built once per worker rather than the thousands of times it was before.

The change is a pure cache: the cached array is built by the identical `np.concatenate`/`np.tile` call, and `np.diag` — its only consumer — returns a read-only view, so the cache cannot be mutated downstream. Model output is bit-for-bit identical to master.

For evidence, a single-reform TPI run on OG-ETH calibration with 7 workers, measured under py-spy (overhead is present on both sides, so the ratio is the meaningful figure):

| | reform wallclock |
|---|---|
| master | 1398.1s |
| this PR | 472.9s |

That is a 2.96x speedup. Every steady-state and transition-path variable in the output — 139 of them — is bit-for-bit identical to master. `tests/test_household.py` passes (51 tests, including `test_FOC_savings` and `test_FOC_labor`), and the full non-`local` suite passes (534 tests).

cc @jdebacker @rickecon

---

**Edit:** I've folded a second speedup and the requested version/lock updates into this PR.

### A second speedup, on top of the e_long fix

Once `e_long` was cached, profiling pointed at the next bottleneck: the tax rate calculation. Every time the model computes tax rates, it was rebuilding the same parameter arrays from scratch — work that's identical from one call to the next.

The fix builds those arrays once and reuses them. Same idea as the `e_long` change, just a different spot. It's worth about another 10%.

Two things worth noting:
- Results are still bit-for-bit identical to master.
- The `mono` and `mono2D` tax functions work differently — they store functions, not numbers — so the fix skips them and leaves their path untouched. A new test covers all seven tax function types.

### @jdebacker's two requests — done

- [x] Version in `__init__.py` bumped to 0.15.12
- [x] `uv.lock` updated — only the `ogcore` version line changed, nothing else

### Updated numbers

Same single-reform run, with both speedups in:

| | reform wallclock |
|---|---|
| master | 1398.1s |
| e_long fix only | 472.9s |
| both fixes | 424.9s |

That's 2.96x from the first fix, 3.29x with both. Output is still identical to master — all 139 steady-state and transition-path variables match. Full non-`local` test suite passes (541 tests).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.15.12] - 2026-05-14 12:00:00
+
+### Added
+
+- Caches the `e_long` array in `household.FOC_savings` and `household.FOC_labor` rather than rebuilding it on every call. `e_long` is a pure function of `p.e`, `p.S`, and `p.J`, none of which change during a solve, so a `_get_e_long` helper now builds it once per worker and reuses it. Profiling identified this rebuild as the single most expensive operation in a TPI run. The change is a pure cache — model output is bit-for-bit identical to master — and gives roughly a 3x speedup on a single-reform TPI run. See PR [#1128](https://github.com/PSLmodels/OG-Core/pull/1128).
+
 ## [0.15.11] - 2026-05-08 12:00:00
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Caches the `e_long` array in `household.FOC_savings` and `household.FOC_labor` rather than rebuilding it on every call. `e_long` is a pure function of `p.e`, `p.S`, and `p.J`, none of which change during a solve, so a `_get_e_long` helper now builds it once per worker and reuses it. Profiling identified this rebuild as the single most expensive operation in a TPI run. The change is a pure cache — model output is bit-for-bit identical to master — and gives roughly a 3x speedup on a single-reform TPI run. See PR [#1128](https://github.com/PSLmodels/OG-Core/pull/1128).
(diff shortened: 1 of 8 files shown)


jdebacker:
@seacelo Excellent PR!  I ran this branch and confirmed the significant speedup in the model solution.  

Can you add two things to these changes:
- [x] Update the version in `__init__.py`
- [x] Commit any changes to `uv.lock`?

CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
...
+- Builds the per-period tax-parameter slices in `TPI.inner_loop` as numpy arrays via a new `_params_to_array` helper, and switches `txfunc.get_tax_rates` to `np.asarray`, so the repeated per-call list-to-array conversion is skipped on the hot TPI path. `mono` and `mono2D` tax functions store callables rather than numbers, so their nested-list form is passed through unchanged. Profiling identified this conversion as the next hot spot after the `e_long` rebuild. Model output is bit-for-bit identical to master, and the change gives roughly a further 10% speedup on a single-reform TPI run (about 3.3x cumulative versus master). See PR [#1128](https://github.com/PSLmodels/OG-Core/pull/1128).
+
```
jdebacker, line 16: Let's add two bullets to this for other PRs that will be merged before this new release:

* Changes the minimum of the allowable range for `tau_c` in `default_parameters.py` to allow for government consumption subsidies.
* Fixes a bug in the `parameter_plots.plot_fert_rates` function.

marcelolafleur: done

jdebacker:
Great first contribution to OG-Core, @seacelo.  Merging.

PR #1144, opened 2026-06-04
Add sparse FOC Jacobian to the household solve

marcelolafleur:
~2× speedup on the household solve, with no aggregate drift. The change: use the Jacobian's known sparsity instead of building it dense.

At each step, the solver needs to know how every equation reacts when each household choice changes. Today it works through them one at a time, even though almost every equation only depends on a couple of nearby ages. Most of that work rediscovers zeros.

Sparsity of the household equation Jacobian at S=12

Rows are equations, columns are the household's choices (`b` = savings, `n` = labor — one of each per age). Left: how the solver currently sees the problem — assume any equation could react to any choice. Right: reality. Each savings or labor equation only touches a household's own age and an immediate neighbor; the rest is zeros.

Once you know that structure, you can test choices in batches whose equations don't overlap — about seven checks per step instead of 160 at the default S=80. Same answers, roughly half the wall time.

This PR implements that method, currently as an optional flag (`use_sparse_FOC_jac`).

Math derivation in the new `derivations.md` appendix; short note in `equilibrium.md`. New CI test (`tests/test_TPI.py::test_inner_loop_sparse_FOC_jac`) mirrors the existing inner-loop test with the flag on.

`examples/run_sparse_FOC_jac_compare.py` runs a baseline twice (off, then on) and prints a NO DRIFT / DRIFT DETECTED verdict against a 0.1% threshold. No argument runs OG-Core's standard example; passing a country package name (e.g. `ogphl`) runs that country's baseline. Numbers on an M1 MacBook Pro:

| | dense | sparse | speedup | worst aggregate drift |
|---|---|---|---|---|
| OG-Core (standard example) | 385 s | 188 s | 2.05× | 1.88e-12 |
| OG-ETH | 284 s | 149 s | 1.92× | 4.73e-11 |
| OG-ZAF | 450 s | 196 s | 2.30× | 0.069% |
| OG-PHL | 525 s | 222 s | 2.36× | 6.58e-09 |
| OG-IDN | 319 s | 160 s | 2.00× | 3.30e-13 |

All five pass the 0.1% threshold at baseline. Reforms not tested.

ZAF's 0.069% drift is concentrated entirely on aggregate bequests; every other ZAF aggregate matches at machine precision. The cause traces to ZAF's calibrated productivity profile for the lowest income group, which drops 10–30× below other countries' at older ages. At that level, labor income is a rounding error in the household's budget, so for a handful of cohorts there are several equally-good lifetime work-and-save patterns. Dense and sparse converge to different valid patterns for these cohorts. Verified by substituting OG-Core's productivity profile for the lowest income group, with ZAF's everything else: the drift disappears completely. This is a property of the ZAF calibration, not a solver issue.

cc @jdebacker @rickecon


jdebacker (dismissed):
@SeaCelo These changes look very good to me.  Let me replicate your tests and confirm back here.  

Before we merge, I think we'll want to put out the `run_sparse` example script and turn on the use of the sparse Jacobian in the default parameters.    But it's good to have those the way they are for now to enable a bit more testing.

Great job identifying a way to improve the numerical optimization.

markdebacker:
@SeaCelo I ran your test script on a windows machine with 32 gb of ram and a 12th Gen Intel(R) Core(TM) i5-12600KF  processor (with 10 cores). Here are the results: 
```
  MODEL:  ogcore (standard example)
  SPEED
    dense  :   740.12 s
    sparse :   369.18 s   ->   2.00x faster

  DRIFT  (max relative difference, sparse vs dense)
    TPI worst:  BQ                        0.0000%   (8.00e-12)
    SS  worst:  B                         0.0000%   (0.00e+00)

    All TPI aggregates, sorted by drift:
      BQ                        0.0000%   (8.00e-12)
      I_total                   0.0000%   (3.61e-12)
      C                         0.0000%   (1.69e-12)
      r_gov                     0.0000%   (5.62e-13)
      r_p                       0.0000%   (3.87e-13)
      r                         0.0000%   (3.75e-13)
      K                         0.0000%   (2.72e-13)
      B                         0.0000%   (2.62e-13)
      total_tax_revenue         0.0000%   (1.58e-13)
      L                         0.0000%   (1.25e-13)
      w                         0.0000%   (1.07e-13)
      Y                         0.0000%   (9.69e-14)
      TR                        0.0000%   (8.68e-14)
      D                         0.0000%   (4.16e-14)

  ACCURACY FLOOR  (resource-constraint residual)
    dense  :  8.46e-07
    sparse :  8.46e-07

  RESULT:  NO DRIFT  (worst 0.0000%  <=  0.1% threshold)
```

rickecon:
YESSS! Not only is @SeaCelo turning out to have the most epic performance upgrades to OG-Core, but @markdebacker is on GitHub and running tests on OG-Core. We've got to get @markdebacker to make a PR and be a contributor to this repo.

@SeaCelo. I love the update in this PR. I will spend some time testing it. But I think this is a critical addition to the multivariate root finders in `SS.py` and `TPI.py`. I've used this approach in trying to come up with equilibrium solution algorithms to large scale problems like this.

TabA3-1

marcelolafleur:
Nice to meet you @markdebacker!

jdebacker:
@SeaCelo I'm ready to merge this after you bump the version (to 0.16.1), remove the test script, and make the default parameterization one which uses the sparse Jacobian.

marcelolafleur:
@jdebacker I've rebased this, made it default, removed the test script, and changed the version. I also added a small message in case the solver falls back to dense. Once the tests clear it should be ready to merge. I didn't add the option `use_sparse_FOC_jac=True/False` to the script. The default will just run the sparse method.

jdebacker (approved):

jdebacker:
Great PR. Merging.

PR #1145, opened 2026-06-06
Fix J=1 shape collapse bugs (#1143)

### LINKED ISSUE #1143: Potential issue with `J=1` parameterization
The test `tests/test_TPI.py::test_run_TPI_extra[J=1]` seems to hang -- I haven't had this complete.

I identified the issue when running the full test suite for PR #1126, but the problem exists on `master` without those changes.

My hunch is that this may be related to changes in PR #1128, but I am not sure.

### LINKED ISSUE #1146: Sweep remaining np.squeeze J=1 shape-collapse bugs across tax functions and TPI paths
Follow-up to #1143 / #1145.

PR #1145 fixed three sites of a shape-collapse pattern that broke J=1. The same pattern is in more places — they aren't exercised by the testing baseline (which uses linear tax functions) but they will bite J=1 users on other tax function types or in TPI.

The pattern is `np.squeeze` applied to an (S, J) array. For J≥2 it does nothing. For J=1 it collapses the J axis down to a flat (S,) list. The flat list then gets multiplied by something else of shape (S, J) or (S, 1), and numpy broadcasts that into an (S, S) grid. Summing the grid adds up S² numbers instead of S, so the result is S times too big.

Places I know have this:

- `txfunc.get_tax_rates` — the GS, HSV, DEP, DEP_totalinc, and mono branches all have the same `np.squeeze(params[..., i])` pattern. #1145 only fixed the linear branch.
- `aggregates.py` TPI branches of `get_L`, `get_B`, `get_I`, `get_BQ`, `get_C` — each uses `np.squeeze(p.lambdas)` against (T, S, J) shapes.
- `household.py` `FOC_savings`, `FOC_labor`, `get_y` — the j=None paths squeeze `p.e[-1, :, :]`.
- A couple more in `SS.py` — most are squeeze-then-reshape which is safe, but a few aren't.

One thing to watch when fixing these: the same syntax means different things depending on the caller. In the aggregate (j=None) SS path, `params` is (S, J, N) and `X` is (S, J), and the squeeze breaks J=1. In the per-j SS path, `params` is the same shape but `X` is just (S,), and the squeeze is actually needed to collapse J=1 down to match. So you can't just delete every `np.squeeze` — you need a shape-aware extraction. The fix #1145 used in the linear branch is the template:

```python
rate = np.asarray(params[..., 0])
if rate.size == np.size(income):
    rate = rate.reshape(np.shape(income))
else:
    rate = np.squeeze(rate)
```

For evidence: when I traced the linear-branch site during a J=1 SS run, `income_tax_liab` was getting `b=(40, 1)`, `e=(40, 1)`, `n=(40, 1)`, but `ETR_income`'s output came back (40, 40) — a clean outer product from `phi0 * income` after `phi0` got collapsed to (40,) by the squeeze. The same trace would catch each other site if you exercise it with a J=1 calibration on that tax function type.

Each fix should come with a regression test on its function (J=1 aggregate-shape inputs → output shape matches input) and the existing per-j tests for J=2 should keep passing.

marcelolafleur:
Fixes #1143. Closes #1146.

OG-Core is almost always run with several income groups, and a handful of spots assumed there was always more than one. Each used `np.squeeze`, which drops a size-1 axis — a no-op for J≥2, but for J=1 it collapses the `(S, 1)` ability column to a flat `(S,)`. When that flat array is then multiplied against an `(S, 1)` array, numpy builds an `(S, S)` outer product instead of multiplying element by element, so the result comes out S× too large or the wrong shape. For J=1 this broke aggregate labor, the income-scaling factor, tax rates, and before-tax income — enough that the steady state couldn't clear (K explodes, r and consumption go negative).

Sites fixed:
- `aggregates.get_L` — SS labor aggregate
- `SS.py` `average_income_model` — calibration factor
- `txfunc.get_tax_rates` — all parametric tax functions (linear, DEP, DEP_totalinc, GS, HSV), via one shape-aware coefficient helper
- `household.get_y` — before-tax income
- `SS.py` — the `e` argument to the income-tax and pension calls

Why it took a couple of passes: the J=1 test uses a flat productivity profile and the simplest (linear) tax function — settings that happen to hide this bug (an inflated table of ones is still ones), and it never exercises the realistic tax functions. So the test passing wasn't enough; the realistic-tax and before-tax-income sites only surfaced in a full model run. The test also compared against a saved fixture generated under the old buggy code, so it kept failing even after the code was right. That fixture is regenerated.

Validation:
- Full SS + TPI with one income group vs an identical two-group economy (which must give the same answer): match to machine precision across baseline, reform, and use_zeta.
- `test_run_TPI_extra[J=1]` passes; full suite green.

Regression tests:
- `test_get_L_J1_regression` — `get_L` against a plain double-sum reference.
- `test_get_tax_rates_J1_shape` — every parametric tax function preserves the J=1 shape (fails without the fix).
- `test_get_y` strengthened to use realistic `(S, J)` inputs and assert the shape.

Regenerated `inner_loop_outputs_J1.pkl` and `run_TPI_outputs_J1.pkl`, which encoded the old buggy output.

cc: @jdebacker @rickecon @utsinboots


utsinboots:
Tested PR #1145, commit `0ca9da89` on Windows

Passed:
`uv run pytest tests/test_aggregates.py::test_get_L_J1_regression -vv`
`uv run pytest tests/test_txfunc.py::test_get_tax_rates_linear_J1_shape -vv`
`uv run pytest "tests/test_SS.py::test_inner_loop[J=1]" -vv`
`uv run pytest tests/test_aggregates.py` 42 passed

Failed:
`uv run pytest "tests/test_TPI.py::test_run_TPI_extra[J=1]" -vv` 1 failed in 34.39s
Error: `FAILED tests/test_TPI.py::test_run_TPI_extra[J=1] - assert False
 where False = <function allclose at 0x00000299CE1BBBF0>(array([0.76815272, 0.76666378, 0.77806494, 0.77931536, 0.78144139,\n 0.78393469, 0.78673022, 0.78957472, 0.79246368, 0.79573516, <- - - - - -> \n 0.32000643, 0.32000303, 0.3199997 , 0.31999642, 0.31999319]), rtol=0.0001, atol=0.0001)
 where <function allclose at 0x00000299CE1BBBF0> = np.allclose`

Also, `uv run pytest tests/test_txfunc.py` 30 passed, 1 failed
Failure was `tests/test_txfunc.py::test_tax_func_estimate` while reading `tests/test_io_data/tax_func_estimate_inputs.pkl` with error: `ValueError: could not convert string to int`

The remaining `test_run_TPI_extra[J=1]` failure may be related to additional J=1 shape-collapse cases that were intentionally left out of scope for #1145

jdebacker (approved):

jdebacker:
@SeaCelo Thanks for looking into this.  I think the code changes you suggest make sense, but I'm still getting the failure identified in Issue #1143:

```
uv run pytest tests/test_TPI.py::test
_run_TPI_extra
```

Returns:

```
============================= test session starts ==============================
platform darwin -- Python 3.11.13, pytest-9.0.2, pluggy-1.6.0
rootdir: /Users/jason.debacker/repos/OG-Core
configfile: pyproject.toml
plugins: anyio-4.12.1, xdist-3.8.0, cov-7.0.0
collected 8 items

tests/test_TPI.py .......F                                               [100%]

=================================== FAILURES ===================================
___________________________ test_run_TPI_extra[J=1] ____________________________
...
FAILED tests/test_TPI.py::test_run_TPI_extra[J=1] - assert False
```

marcelolafleur:
@jdebacker @utsinboots Thanks both for checking this. 
The first pass left two gaps. The saved comparison file the test checks against was generated under the old buggy code, so the test was comparing correct new numbers against old wrong ones (regenerated now). And the same bug lived in a few spots the test doesn't exercise (the realistic tax functions and before-tax income) which only showed up in a full run.

Fixed now. I ran the full model (SS + TPI) for one income group against an identical two-group economy and they match. I also confirmed the steady state matches across reform and use_zeta. test_run_TPI_extra[J=1] passes and the full suite is green.

Worth confirming on your end with a full J=1 run. the test alone uses simplified settings that hide this class of bug.

utsinboots:
Tested commit `b25ca89` `Windows`
`uv run pytest "tests/test_TPI.py::test_run_TPI_extra[J=1]" -vv` 1 passed in 17.83s
Full TPI suite: `uv run pytest tests/test_TPI.py` 37 passed, 8 warnings in 5652.14s (1:34:12)
This confirms the `test_run_TPI_extra[J=1]` passes in the latest commit.

jdebacker (approved):

jdebacker:
Excellent find and fix. Thanks @SeaCelo.

PR #1149, opened 2026-06-11
Validate beta_annual and chi_b length against the number of income groups

marcelolafleur:
Part of the #1146 family, on the parameter side — the solver fixes in #1145 don't reach it.

`beta_annual` and `chi_b` carry one value per income group, but nothing checks the length users pass. Too few values crash much later (an IndexError deep in the SS solve, far from the cause). Too many run silently using the first J entries — wrong results, no symptom.

This PR checks the lengths right where J is set. One value applies to all groups. J values pass through. A uniform list of the wrong length is reshaped — it carries no extra information, and smaller-J runs that keep the uniform defaults depend on it. A non-uniform mismatch raises immediately, naming the parameter and the expected length.

cc: @jdebacker @rickecon

### DIFF
--- a/ogcore/parameters.py
+++ b/ogcore/parameters.py
@@ -87,6 +87,28 @@ class Specifications(paramtools.Parameters):
         self.T = int(self.T)
         self.J = len(self.lambdas)
 
+        # beta_annual and chi_b carry one value per lifetime-income group, but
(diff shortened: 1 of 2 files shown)


jdebacker:
Good add and consistent with how we extrapolate other parameters. Merging.

PR #1151, opened 2026-06-11
Make get_un_data robust to pandas 3.0 and Bearer-prefixed tokens

### LINKED ISSUE #1150: get_un_data fails on pandas 3.0 (int assigned to string age column)
`ogcore.demographics.get_un_data` raises a `TypeError` on pandas 3.0 when parsing UN World Population Prospects data, so any live UN pull fails.

### What happens

At `ogcore/demographics.py`:

```python
df.loc[df.age == "100+", "age"] = 100
df.age = df.age.astype(int)
```

```
TypeError: Invalid value '100' for dtype 'str'. ... got 'int' instead.
```

### Why

pandas 3.0 makes the new string dtype the default for inferred text columns (`future.infer_string`). The `age` column is parsed as text (`"0"`…`"99"`, `"100+"`), and a strict string column rejects the integer `100`. Pre-3.0 the column was `object` dtype, which tolerated it — so this worked for years and breaks under the currently resolved pandas (3.0.x).

There is a subtlety: the dtype varies by series. Fertility ages (15–49) contain no `"100+"`, so that column parses as `int64`; mortality and population include `"100+"` and parse as `str`. So neither a plain int nor a plain string assignment works for both.

### Scope

Affects every live UN pull — `get_fert`, `get_mort`, `get_imm_rates`, `get_pop_objs`, and `update_from_api=True` calibration. Offline/cached runs that read JSON are unaffected.

### Fix

Normalize the column to `str` before the replacement, then cast:

```python
df.age = df.age.astype(str)
df.loc[df.age == "100+", "age"] = "100"
df.age = df.age.astype(int)
```

Back-compatible with pandas 2.x. Related: `get_un_data` builds the auth header as `"Bearer " + token`, so a `un_api_token.txt` already containing a `Bearer ` prefix yields `"Bearer Bearer …"` and silent auth failure — proposing to harden that in the same PR.

marcelolafleur:
Fixes #1150.

Two robustness fixes to `ogcore.demographics.get_un_data`:

1. **pandas ≥3.0 compatibility.** The `"100+"` top age bin was replaced with the integer `100` in the parsed `age` column, which raises `TypeError` under pandas 3.0's strict string dtype. The dtype also varies by series — fertility (ages 15–49) parses as `int64` with no `"100+"`, while mortality/population parse as `str` and include `"100+"` — so neither a plain int nor a plain string assignment works for both. Fix: normalize the column to `str`, replace `"100+"` → `"100"`, then `astype(int)`. Back-compatible with pandas 2.x; returned data unchanged.

2. **Robust Bearer token.** The auth header was built as `"Bearer " + token`, so a `un_api_token.txt` that already includes a `Bearer ` prefix produced `"Bearer Bearer …"` and silent auth failure (it fell back to the GitHub mirror). An optional leading `Bearer ` is now stripped, so the file may hold either the bare token or the `Bearer …` form the UN portal provides.

Verified on pandas 3.0.3: a live UN pull for Ethiopia (231) succeeds for fertility, mortality, and population (covering both the `int64` and `100+`-string age columns) and authenticates with a `Bearer `-prefixed token (no mirror fallback).

### DIFF
--- a/ogcore/demographics.py
+++ b/ogcore/demographics.py
@@ -82,6 +82,11 @@ def get_un_data(
 
     # get data from url
     payload = {}
+    # Accept a token with or without a leading "Bearer " prefix so the
+    # header isn't doubled into "Bearer Bearer <token>".
+    UN_TOKEN = UN_TOKEN.strip()
+    if UN_TOKEN.lower().startswith("bearer "):
+        UN_TOKEN = UN_TOKEN[len("bearer ") :].strip()
     headers = {"Authorization": "Bearer " + UN_TOKEN}
     response = get_legacy_session().get(target, headers=headers, data=payload)
     # Check if the request was successful before processing
@@ -97,7 +102,12 @@ def get_un_data(
             axis=1,
             inplace=True,
         )
-        df.loc[df.age == "100+", "age"] = 100
+        # The "100+" top age bin appears only in some series (mortality,
+        # population), so the age column may parse as int (no "100+") or
+        # str. Normalize to str so the replacement works under pandas
+        # >=3.0's strict string dtype, then cast the column to int.
+        df.age = df.age.astype(str)
+        df.loc[df.age == "100+", "age"] = "100"
         df.age = df.age.astype(int)
         df.year = df.year.astype(int)
         df = df[df.age < 100]  # need to drop 100+ age category


jdebacker:
Thanks @SeaCelo. Merging.

PR #1142, opened 2026-05-21
Add universal uv-based installer for OG-Core and country repos

marcelolafleur:
This adds a from-scratch installer that takes a user from a bare machine (only `git` required) to a working OG model environment: it installs uv, clones the repo you pick, runs `uv sync --extra dev`, and verifies the import.

* `scripts/install.sh` (macOS/Linux) and `scripts/install.ps1` (Windows) are universal — choose the repo from a menu or pass `--repo` / `-Repo`. The catalog has `og-core` and `og-eth` today; more country calibrations get added as they migrate to uv.
* `scripts/QUICK_INSTALL.md` documents both ways to run it (one-line, or download-and-run), per platform.

cc: @rickecon @jdebacker

PR #1162, opened 2026-06-24
Installer: multi-repo install, catalog expansion, and a machine-readable catalog export

marcelolafleur:
Builds on the universal installer (#1142). Three additions:

**Install several repos in one run.** `--repo` can be repeated (`--repo og-zaf --repo og-idn`) or take a comma list; `--all` installs the whole catalog. (On Windows, where a parameter can't repeat, use the comma form: `-Repo og-zaf,og-idn`.) Each repo clones into its own subfolder under `--dest` and installs independently — if one fails it's reported and the rest continue, with a per-repo summary at the end. `uv` is installed once. A single repo keeps the existing per-step prompts; multiple run as a batch after one confirmation. `--branch` and `--repo-url` stay single-repo only.

**Expanded catalog.** Adds `og-zaf`, `og-idn`, `og-phl` alongside `og-core`/`og-eth`, now that they're uv-native.

**A reliable way to read the catalog.** `--list` (table) and `--list-json` (JSON) print the catalog and exit, and `scripts/repos.json` publishes the same data as a static file — so downstream tooling can stay in sync by fetching one URL instead of scraping the script or `--help`. A small CI check (`.github/workflows/check_catalog.yml`) fails if `install.sh`, `install.ps1`, and `repos.json` ever drift apart.

Hands-free, unattended use: `bash install.sh --all --dest ~/OG --yes`. `scripts/QUICK_INSTALL.md` is updated for all of the above.

Also tightens a few edges: clean error messages for unknown or missing flag values, and a correct non-zero exit when a non-interactive run can't proceed.

**Tested:** every selection mode on bash 3.2 (the macOS default) — repeated / comma / `--all` / dedup / single, plus the guardrail errors — all stopping before any real install; the catalog stays in sync via the new CI check. The PowerShell side mirrors the bash logic and its `-ListJson` path runs in CI, but I haven't yet run the full multi-repo install loop on Windows.

cc: @rickecon @jdebacker

### DIFF
--- a/.github/workflows/check_catalog.yml
+++ b/.github/workflows/check_catalog.yml
@@ -1,0 +1,35 @@
+name: Check installer catalog
+
+on:
+  push:
+    paths:
+      - 'scripts/install.sh'
+      - 'scripts/install.ps1'
+      - 'scripts/repos.json'
+      - '.github/workflows/check_catalog.yml'
+  pull_request:
+    paths:
+      - 'scripts/install.sh'
+      - 'scripts/install.ps1'
+      - 'scripts/repos.json'
+      - '.github/workflows/check_catalog.yml'
+
+jobs:
+  catalog-in-sync:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - name: Emit catalog from each installer
+        run: |
+          bash scripts/install.sh --list-json > /tmp/catalog_sh.json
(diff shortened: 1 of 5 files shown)


jdebacker (approved):

jdebacker:
Nice updates, @SeaCelo.  This will make it simple to add additional country calibrations in the future.

PR #1166, opened 2026-07-07
Extend alpha_FA to the model time path

marcelolafleur:
`alpha_FA` was added in 0.16.0 but never added to the list of parameters that get extended over the time path, so unlike `alpha_G`/`alpha_T`/`alpha_I` you can't give it a multi-year path. A single value works by accident (it broadcasts); any real path stays short and breaks the transition solve when the fiscal code multiplies it by `Y[:T]`.

Fix is to register it in `tp_param_list` so it extends to `T+S` with the last value carried forward. Nothing else needed — SS and TPI already index it like `alpha_G` (`[-1]` for the steady state, `[:T]` on the path). Also updated its description in `default_parameters.json` to note it can vary over time, and added a test.

Scalar `alpha_FA` gives a constant path, so existing runs are unchanged. `test_parameters.py` and `test_fiscal.py` pass, ruff clean.

cc: @jdebacker @rickecon

### DIFF
--- a/ogcore/default_parameters.json
+++ b/ogcore/default_parameters.json
@@ -956,9 +956,9 @@
     },
     "alpha_FA": {
         "title": "Foreign aid payments to domestic government as a share of GDP",
-        "description": "Foreign aid payments to domestic government as a share of GDP.",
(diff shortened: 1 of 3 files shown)


jdebacker (approved):

jdebacker:
@marcelolafleur Thanks for noting this and making `alpha_FA` flexible.

Issue #1014, opened 2025-03-20
Defined Benefits pension system not handling new object sizes


jdebacker:
The defined benefits system functions have not been updated for at least two time varying parameters: earnings ability and retirement age.

These functions should be updated and tested to work with these new dimensions.

The bug here also suggests new tests: SS and TPI tests should include model runs with each pension system.

marcelolafleur:
PR #1167 fixes the crash-level problems in the DB and PS systems (array/scalar mismatches with a real Specifications object) and makes the DB system solve the steady state and the transition path. Mapping what remains open here after that PR:

- Time-varying retirement age in the DB/NDC/PS formulas — the PR uses the steady-state value throughout.
- Time-varying earnings profiles in the same formulas — the PR uses the steady-state profile.
- Pre-time-path histories: the PR anchors pre-period-0 wages to the period-0 wage of the current path and takes labor from the model's initial condition. A fully history-consistent treatment (each cohort's own solved pre-path labor) is still open.
- TPI-level model-run tests for each pension system (the PR adds a steady-state solve test for DB).

Downstream context: OG-BRA now runs the DB system for Brazil's RGPS, so these have a live user.

PR #1167, opened 2026-07-09
Fix the Defined Benefits and Points System pensions so they can actually run

### LINKED ISSUE #1075: Bug: Points System crashes with ValueError due to p.retire being a list (plus subsequent TypingError)
Hi, 

I am attempting to run the model using the "Points System" for pensions. I encountered a blocking error immediately when running the baseline.

1. The Primary Issue (ValueError) When p.retire is a list (representing the time path, length 400), the function PS_amount in pensions.py crashes at np.zeros(p.retire). It interprets the list as dimensions rather than the length of the array.

Traceback:
File "C:\...\ogcore\pensions.py", line 412, in PS_amount
    L_inc_avg_s = np.zeros(p.retire)
ValueError: maximum supported dimension for an ndarray is currently 64, found 400

2. Secondary Issue (TypingError) If I manually patch the code above to use int(p.retire[0]), I immediately hit a second error regarding g_y (growth rate). If g_y is a scalar (float), the Numba-compiled loop fails because it tries to access g_y[-1].

Traceback (after patching the ValueError):
numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
No implementation of function Function(<built-in function getitem>) found for signature:
 >>> getitem(float64, int64)
...
File "...\ogcore\pensions.py", line 888:
        for s in range(S_ret):
            L_inc_avg_s[s] = w[s] / np.exp(g_y[-1] * (u - s)) * e[s] * n[s]
            ^

Environment:
- OS: Windows 11
- Python 3.11 / Anaconda

Thanks for your help!

marcelolafleur:
The non-US pension systems in `pensions.py` crash when used with a real Specifications object. We found this calibrating OG-BRA: Brazil's pension system (RGPS) is a defined benefits scheme, so OG-BRA is the first country model to use `pension_system = "Defined Benefits"` — and it crashes on the first solve. Same family of bugs as the Points System crash reported in #1075.

The cause is that these code paths were never updated after some parameters became time-varying: the functions pass arrays where the numba loops expect scalars, and scalars where they expect arrays. The existing tests didn't catch this because they feed the loops pre-cleaned inputs from mock objects instead of a real Specifications object.

Three changes (the third was added after testing a full transition path downstream):

- Fix the Defined Benefits and Points System paths so they run. They use the steady-state retirement age and earnings profile for now; full time variation stays open in #1014. The NDC system needs more than this — its growth-rate settings aren't in `default_parameters.json` yet — so it remains unusable for now.
- Add tests that call the pension functions with a real Specifications object (these fail on master with the #1075 errors), plus a steady-state solve test with the DB system.
- Make the DB system solve the transition path, not just the steady state. Two pieces were missing: the benefit formulas reference pre-time-path wages and labor that nothing ever supplied (labor now comes from the model's initial condition — the same baseline object that initializes wealth — and wages are anchored to the period-0 wage of the current path), and there was no evaluation of DB amounts over the full (T, S, J) path for the TPI's aggregate revenue step (added as `DB_3dim_loop`, built from each cohort's own wage history so aggregates match household behavior; it reproduces the steady-state solution to machine precision on a constant path).

Fixes #1075. Part of #1014.

Tested: full pension and tax suites pass (71 tests); the DB steady state solves. Downstream check: the OG-BRA Defined Benefits calibration solves both the baseline and a reform transition path against this branch, passes the resource-constraint check on both, and hits its pension spending target to four decimals.

cc @rickecon @jdebacker

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,11 +21,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   growth-rate settings (`ndc_growth_rate`, `dir_growth_rate`) that are
   not yet parameters in `default_parameters.json`, so it still cannot
   run against a real `Specifications` object.
+- Wired the pre-time-path inputs the DB/NDC/PS benefit formulas need
+  into the TPI: labor supplied before the time path begins comes from
+  the model's initial condition (the same baseline object that
+  initializes wealth), and pre-time-path wages are anchored to the
+  period-0 wage of the current path. Added the full time-path (T x S x J)
+  evaluation of Defined Benefits amounts (`DB_3dim_loop`) used when the
+  TPI computes aggregate revenues, built from each cohort's own wage and
+  labor history so aggregates are consistent with household behavior.
+  With these changes a country model using the Defined Benefits system
+  solves both the steady state and the transition path.
 - Added regression tests that call `pension_amount` with a real
   `Specifications` object per pension system (the existing tests
   pre-scalarized the inputs and so never exercised the real interface)
   and a local-marked steady-state solve test with the Defined Benefits
-  system.
+  system. See PR [#1167](https://github.com/PSLmodels/OG-Core/pull/1167).
 
 ## [0.16.4] - 2026-07-02 12:00:00
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   applied to the NDC path, but the NDC system additionally requires
   growth-rate settings (`ndc_growth_rate`, `dir_growth_rate`) that are
   not yet parameters in `default_parameters.json`, so it still cannot
-  run against a real `Specifications` object.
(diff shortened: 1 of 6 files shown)


jdebacker:
@marcelolafleur I appreciate the clarity in this PR about what is addressed and what is not.  While not totally solving all of Issue #1014, I think this PR is an important addition and fix of the DB and PS pension systems.  I want to merge this and create a new release.  Can you update the version to 0.17.0 and include in the change log not just this PR, but the changes in PRs #1174, #1171, and #1166?

marcelolafleur:
Thanks @jdebacker. All set:

Version bumped to 0.17.0 (pyproject.toml and ogcore/__init__.py).
New [0.17.0] changelog section covering this PR plus [#1174](https://github.com/PSLmodels/OG-Core/pull/1174), [#1171](https://github.com/PSLmodels/OG-Core/pull/1171), and [#1166](https://github.com/PSLmodels/OG-Core/pull/1166).
Also corrected a stale docstring in DB_3dim_loop that said the function wasn't jitted — it is in the current version of numba.

Ready to merge whenever you are.

PR #1171, opened 2026-07-10
Use capital income tax noncompliance rate in SS etr/mtry diagnostics

### LINKED ISSUE #1170: SS output "etr" and "mtry" use the labor noncompliance rate in place of the capital rate
In `SS.py` (~line 915 on current master), the capital noncompliance array is tiled from the wrong parameter — a copy-paste of the labor block above it:

```python
capital_noncompliance_rate_2D = np.tile(
    np.reshape(p.labor_income_tax_noncompliance_rate[-1, :], (1, p.J)),
    (p.S, 1),
)
```

It should use `p.capital_income_tax_noncompliance_rate` (as `TPI.py` ~line 1468 does for the equivalent 3D array).

This is a reporting bug, not a solution bug. The array only feeds the post-solve `etr_ss` and `mtry_ss` diagnostics, so the `"etr"` and `"mtry"` entries of the SS output are wrong whenever the two noncompliance rates differ. The household FOCs and tax liabilities go through `household.py` and `tax.py`, which index the capital rate correctly, so the equilibrium and all aggregates are unaffected.

Dates to #816. Related: #1168. One-word fix; I'll open a small PR.

marcelolafleur:
Fixes #1170.

One-word fix: `capital_noncompliance_rate_2D` in `SS.py` was tiled from `p.labor_income_tax_noncompliance_rate` instead of `p.capital_income_tax_noncompliance_rate`. This only affects the post-solve `etr_ss` and `mtry_ss` diagnostics in the SS output — the solution itself uses the correct rates via `household.py` and `tax.py`. No behavior change when the two rates are equal (the default).

Happy to add a changelog entry and version bump if you'd like them in this PR.

### DIFF
--- a/ogcore/SS.py
+++ b/ogcore/SS.py
@@ -913,7 +913,7 @@ def SS_solver(
         (p.S, 1),
     )
     capital_noncompliance_rate_2D = np.tile(
-        np.reshape(p.labor_income_tax_noncompliance_rate[-1, :], (1, p.J)),
+        np.reshape(p.capital_income_tax_noncompliance_rate[-1, :], (1, p.J)),
         (p.S, 1),
     )
     income_tax_filer_2D = np.tile(


jdebacker (approved):

jdebacker:
@marcelolafleur Good catch. Merging.  We'll wrap this into the next release.

PR #1174, opened 2026-07-15
Use each year's tax noncompliance and filer rates over time

### LINKED ISSUE #1168: TPI applies year-0 tax noncompliance and filer rates to the whole transition path
On the transition path, `income_tax_liab` picks a single row of the noncompliance and filer arrays:

```python
tax_filer = p.income_tax_filer[t, :]
```

and the path-wide callers pass `t=0` — `TPI.py` (~lines 1046 and 1064) and `aggregates.revenue` (~line 397). So if these parameters vary over time, households respond to the changing rates, but revenue, the government budget, and the debt path all use the year-0 values. In a country calibration where noncompliance falls over five years, labor supply responds but income tax revenue as a share of GDP stays exactly at baseline.

`tau_payroll` a few lines below in the same function is sliced correctly (`p.tau_payroll[t : t + length]`). The fix is probably to slice these arrays the same way, shaped `(T, 1, J)`. The `j` branch of the function has the same pattern and should get the same treatment.

Verified on current master (`a12ecf5`, v0.16.3). I'll create a PR for this.

marcelolafleur:
Fixes #1168.

The model lets you set noncompliance and tax-filer rates that change from year to year. But one of the revenue calculations ignored those changes and used the first year's rates for every year. So households reacted to the changing rates while the government's tax revenue, budget, and debt stayed stuck on the first-year numbers — the two sides of the model disagreed.

This makes that calculation use each year's actual rates, the same way the payroll tax rate right beside it already does.

Added two tests that check the year-by-year tax when the rates change over time; they fail without the fix. Nothing changes when the rates stay constant, which is the default.

### DIFF
--- a/ogcore/tax.py
+++ b/ogcore/tax.py
@@ -338,16 +338,39 @@ def income_tax_liab(r, w, b, n, factor, t, j, method, e, etr_params, p):
     """
     if j is not None:
         if method == "TPI":
+            length = r.shape[0]
             if b.ndim == 2:
                 r = r.reshape(r.shape[0], 1)
                 w = w.reshape(w.shape[0], 1)
+                labor_income_tax_compliance_rate = (
+                    p.labor_income_tax_noncompliance_rate[
+                        t : t + length, j
+                    ].reshape(length, 1)
+                )
+                capital_income_tax_compliance_rate = (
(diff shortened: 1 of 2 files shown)


jdebacker:
@marcelolafleur Good bug fix. Merging.

PR #1182, opened 2026-07-23
Default income_percentiles to a single group in the homogeneous case

### LINKED ISSUE #1180: 0.18.0: get_pop_objs requires income_percentiles even in the homogeneous case (assert precedes the all-None branch)
In 0.18.0, `get_pop_objs` fails for the homogeneous (no income heterogeneity) case that 0.17.x supported, because `expand_pop_obj_J` asserts `income_percentiles is not None` *before* reaching its own all-inputs-`None` branch.

## Repro

```python
from ogcore import demographics

demographics.get_pop_objs(
    E=20, S=80, T=320,
    country_id="826",
    initial_data_year=2026,
    final_data_year=2027,
    GraphDiag=False,
)
```

```
File "ogcore/demographics.py", line 1591, in get_pop_objs
    pop_objs = expand_pop_obj_J(
File "ogcore/demographics.py", line 942, in expand_pop_obj_J
    assert income_percentiles is not None, (
AssertionError: income_percentiles must be provided when using income-specific inputs.
```

## Cause

`get_pop_objs` now calls `expand_pop_obj_J` unconditionally, and in `expand_pop_obj_J` the assert at demographics.py:942 sits above the `all_income_inputs_none` early-return branch (~:953) that handles the homogeneous case. So the assert fires even when every income-specific input (`fert_gradient`, `mort_gradient`, `infmort_gradient`, `imm_pctiles`) is `None` — i.e. for every pre-0.18 call signature. `income_percentiles` defaults to `None` in `get_pop_objs`, which makes the default signature self-contradictory.

## Suggested fix

Either default to a single income group when all income-specific inputs are `None` (e.g. treat `income_percentiles=None` as `[100]` in that branch, preserving the 0.17 behavior and output shapes for existing consumers), or move the assert inside the income-specific path — plus a release-note flag if requiring `income_percentiles` is intended, since it's a breaking change for every downstream caller.

## Downstream impact

OG-UK's `calibrate()` breaks on 0.18.0; we capped `ogcore>=0.15.6,<0.18` in PSLmodels/OG-UK#69 pending resolution here, with migration to the income-heterogeneity API as a follow-up.

(Separate minor note hit while debugging: `get_un_data` wraps its `input()` token prompt in `except EOFError` only, so under pytest's capture the stdin read raises `OSError` instead of falling through — an env-var token path would make headless runs cleaner. Happy to file separately if useful.)

MaxGhenis:
Fixes #1180.

## Problem

In 0.18.0, `get_pop_objs` raises an `AssertionError` for the homogeneous case — the pre-0.18 call signature with `income_percentiles=None` and no income-specific inputs — because `expand_pop_obj_J` asserts `income_percentiles is not None` before reaching its own all-inputs-`None` branch. Every downstream caller using the 0.17 signature breaks (OG-UK capped `ogcore<0.18` in PSLmodels/OG-UK#69 pending this fix).

## Fix

`expand_pop_obj_J` now computes `all_income_inputs_none` first; when `income_percentiles is None` it asserts only that no income-specific inputs were supplied, then defaults to a single income group (`[100]`, J=1), broadcasting the aggregate objects as the homogeneous branch intends. `income_percentiles` remains required whenever gradients or immigrant income shares are provided.

Also included:
- removes a stray debug `print("In the all are none case.")` from the homogeneous branch;
- corrects the `expand_pop_obj_J` docstring — the aggregate objects are broadcast to J=1, not "returned" in their original 2-D shapes (`get_pop_objs` indexes three dimensions unconditionally) — and documents the default in both docstrings;
- CHANGELOG entry and version bump to 0.18.1 (happy to drop the bump if you prefer to handle versioning separately).

## Tests

Two synthetic unit tests alongside the existing `expand_pop_obj_J` tests:
- `test_expand_pop_obj_J_defaults_to_single_group` — the pre-0.18 signature returns `(T, S, 1)` objects equal to the aggregates (regression for #1180);
- `test_expand_pop_obj_J_income_inputs_require_percentiles` — a gradient without percentiles still raises.

```
$ pytest tests/test_demographics.py -k expand -q
4 passed, 18 deselected in 1.17s
```

End-to-end, the exact repro from #1180 now succeeds (UK, 2026–2027, GitHub population-data fallback):

```
omega (400, 80, 1) / omega_SS (80, 1) / rho (400, 80, 1) / imm_rates (400, 80, 1); g_n_ss 0.001054
```

`ruff format --check` and `ruff check` both pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -685,6 +685,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Any earlier versions of OG-USA can be found in the [`OG-Core`](https://github.com/PSLmodels/OG-Core) repository [release history](https://github.com/PSLmodels/OG-Core/releases) from [v.0.6.4](https://github.com/PSLmodels/OG-Core/releases/tag/v0.6.4) (Jul. 20, 2021) or earlier.
 
 
+[0.18.1]: https://github.com/PSLmodels/OG-Core/compare/v0.18.0...v0.18.1
 [0.18.0]: https://github.com/PSLmodels/OG-Core/compare/v0.17.0...v0.18.0
 [0.17.0]: https://github.com/PSLmodels/OG-Core/compare/v0.16.4...v0.17.0
 [0.16.4]: https://github.com/PSLmodels/OG-Core/compare/v0.16.3...v0.16.4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.18.1] - 2026-07-22 12:00:00
+
+### Bug Fixes
+
+- Fixes Issue [#1180](https://github.com/PSLmodels/OG-Core/issues/1180):
+  `get_pop_objs` raised an `AssertionError` for the homogeneous case
+  (`income_percentiles=None` with no income-specific inputs), because
+  `expand_pop_obj_J` asserted before reaching its homogeneous branch.
+  `income_percentiles` now defaults to a single income group (J=1) when
+  no income-specific inputs are supplied, restoring the pre-0.18 call
+  signature; it remains required whenever gradients or immigrant income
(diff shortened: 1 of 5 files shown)


jdebacker (approved):

jdebacker:
@MaxGhenis Thanks for the bug fix!

PR #1183, opened 2026-07-23
Change default `stationarized` to `False` in `plot_aggregates`

### LINKED ISSUE #1133: Change the default `stationarized` in `output_plots.plot_aggregates`
We should change the default `stationarized` in `output_plots.plot_aggregates` to `stationarized=False` to plot the unstationarized percentage changes.  This shouldn't have any different if the underlying growth rates in the baseline and reform are the same and will give a more consistent view of the changes in aggregates if they are not.

arihantlodha-cmd:
Closes #1133.

Changes the default `stationarized` argument in `output_plots.plot_aggregates` from `True` to `False`, so unstationarized percentage changes are plotted by default. As noted in the issue, this produces identical output when the baseline and reform share the same underlying growth rates, and a more consistent view of aggregate changes when they don't.

**Testing**
- `pytest tests/test_output_plots.py` → 50 passed
- `ruff format --check .` and `ruff check .` → clean

One question for maintainers: `plot_industry_aggregates` in the same module also defaults to `stationarized=True`. I left it unchanged since the issue only mentions `plot_aggregates` — happy to flip it here too if you'd like the two kept consistent.

### DIFF
--- a/ogcore/output_plots.py
+++ b/ogcore/output_plots.py
@@ -188,7 +188,7 @@ def plot_industry_aggregates(
     var_list=["Y_m"],
     ind_names_list=None,
     plot_type="pct_diff",
-    stationarized=True,
+    stationarized=False,
     num_years_to_plot=50,
     start_year=DEFAULT_START_YEAR,
     forecast_data=None,
--- a/ogcore/output_plots.py
+++ b/ogcore/output_plots.py
@@ -22,7 +22,7 @@ def plot_aggregates(
     reform_params=None,
     var_list=["Y", "C", "K", "L"],
     plot_type="pct_diff",
-    stationarized=True,
+    stationarized=False,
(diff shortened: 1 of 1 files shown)


jdebacker:
@arihantlodha-cmd: good catch. Yes, please change `plot_industry_aggregates`.

arihantlodha-cmd:
Thanks! Done — also changed the default to `stationarized=False` in `plot_industry_aggregates` for consistency. I also synced the branch with `master`, which clears the lint failure (it was the unrelated `tests/BENCHMARK_README.md` formatting on the stale base). All 50 `test_output_plots.py` tests pass.

jdebacker (approved):

jdebacker:
@arihantlodha-cmd Thanks for the PR.  Merging.

PR #1184, opened 2026-07-23
Add `get_payroll_tax_revenue` function for correct payroll tax revenue

### LINKED ISSUE #1023: Computing payroll tax revenue
Payroll tax revenue is computed in `aggregates.revenue` as:

```python
payroll_tax_revenue = (
            p.frac_tax_payroll[: p.T] * iit_payroll_tax_revenue
        )
```

This implicitly assumes one is using personal income tax functions that include both payroll and income taxes and that the users is separating these revenues in proportion to a series they input in the parameter `frac_tax_payroll` (which they may have found from a microsimulation model of payroll and income taxes).  

The issue is that this calculation will be incorrect when a user includes payroll taxes via `tau_payroll` (and excludes them from the tax functions represented in `etr_params`, `mtrx_params`, and `mtry_y` params.

I think the solution is to create a payroll tax revenue function for this calculation.  This function will then have a flag to determine how the user treated payroll taxes in the model.  e.g., something like:

```python
def payroll_tax_revenue(args...):
    if np.any(tau_payroll != 0):
        payroll_tax_revenue = p.tau_payroll[:T] * w[:T] * L[:T]
    else:
         payroll_tax_revenue = (
                     p.frac_tax_payroll[: p.T] * iit_payroll_tax_revenue
                )
```

arihantlodha-cmd:
Closes #1023.

Adds a dedicated `get_payroll_tax_revenue` function in `aggregates.py` and calls it from `revenue()` in place of the two inline calculations.

Previously payroll tax revenue was always computed as `frac_tax_payroll * iit_payroll_tax_revenue`, which assumes payroll taxes are embedded in the estimated income and payroll tax functions and separated out via the `frac_tax_payroll` series. As noted in the issue, that is incorrect when a user instead models payroll taxes explicitly through `tau_payroll` (and excludes them from `etr_params`/`mtrx_params`/`mtry_params`). The new function branches on `tau_payroll`:

- if `np.any(tau_payroll != 0)`: `payroll_tax_revenue = tau_payroll * w * L` (aggregate labor summed across industries)
- otherwise (the default, `tau_payroll == 0`): the existing `frac_tax_payroll * iit_payroll_tax_revenue`

The two are identical when `tau_payroll == 0`, so default runs are unchanged (existing `test_revenue` still passes).

A few implementation choices, flagged for review:
- Named it `get_payroll_tax_revenue` rather than `payroll_tax_revenue` to match the module's `get_*` convention and to avoid shadowing the local `payroll_tax_revenue` variable inside `revenue()`. Happy to rename if you'd prefer.
- `L` arrives per-industry in `revenue()`, so aggregate labor income uses `L.sum(-1)`.
- Consolidated to a single call after the SS/TPI branches rather than duplicating the calculation in each.

**Testing**
- Added `test_get_payroll_tax_revenue` covering both branches for SS and TPI.
- `pytest tests/test_aggregates.py -m "not local and not benchmark"` → 43 passed; `ruff format --check .` and `ruff check .` clean.

**One question for maintainers:** when payroll taxes are modeled via `tau_payroll` and excluded from the tax functions, `iit_payroll_tax_revenue` no longer contains payroll revenue, so `iit_revenue = iit_payroll_tax_revenue - payroll_tax_revenue` (and payroll's contribution to `total_tax_revenue`) may also need revisiting. I left those as-is to keep this PR scoped to the issue — glad to follow up if you'd like that handled here too.


jdebacker:
@arihantlodha-cmd Thanks for this PR. 

Regarding your question, yes, let's fix the calculation of `iit_revenue`.  I think something like the following could work (placed before the `total_tax_revenue =` line):

```python
payroll_tax_revenue = get_payroll_tax_revenue(
        w, L, iit_payroll_tax_revenue, p, method
    )
if np.any(p.tau_payroll != 0):
    iit_payroll_tax_revenue += payroll_tax_revenue
iit_revenue = iit_payroll_tax_revenue - payroll_tax_revenue
```

jdebacker:
Also, please run `make format` to resolve formatting issues.  Thanks!

arihantlodha-cmd:
Thanks for the review! Addressed both points:

- Added the `iit_revenue` fix as suggested: `iit_payroll_tax_revenue += payroll_tax_revenue` when `np.any(p.tau_payroll != 0)`, so the explicitly-modeled payroll revenue now flows into both `total_tax_revenue` and `iit_revenue`.
- Regenerated the `test_revenue` golden values, since its params use `tau_payroll=0.5`. Each new total equals the previous total plus `payroll_tax_revenue` (verified by construction); default `tau_payroll==0` runs are unchanged.

Ran `make format` and synced with `master`, so the lint check passes now (`ruff format --check .`, `ruff check .`, `linecheck` all clean). Full `test_aggregates.py` passes (43).

jdebacker (approved):

PR #1005, opened 2025-03-09
Model output variables

jdebacker:
This PR addresses Issue #955 by harmonizing names out of the output variables.  

The PR also adds a new `model_variables.json` that contains metadata for each output variables.  These metadata are used to build a chapter in the OG-Core documentation that summaries the output variables and replaces the output variable labels (used in `output_plots.py` and `output_tables.py`) that had bee in `constants.py`.

PR #1185, opened 2026-07-24
Make SS and TPI output dictionaries expose the same keys

### LINKED ISSUE #1152: Make consistent output dictionaries for SS and TPI results
PR #1005 made the names for keys in the output dictionaries with the steady-state and transition path results consistent.  But it did not ensure that all the same keys are available in each dictionary.  For example "labor_income" is in the transition path output, but not in the steady-state output.  It would be more intuitive for the user if all of the same keys were available in each set of output.

arihantlodha-cmd:
Closes #1152.

PR #1005 made the key *names* in the steady-state and transition-path output dictionaries consistent, but the two dictionaries still exposed different sets of keys. This makes the same keys available in both.

Before this PR the differences were:

- In the TPI output but missing from SS: `labor_income`, `capital_income`, `income_payroll_taxes`
- In the SS output but missing from TPI: `factor`, `theta`

Changes:

**SS (`ogcore/SS.py`)**
- Add `labor_income` (`w * n * e`) and `capital_income` (`r_p * b_s`) to the output dict, computed the same way as their TPI counterparts.
- Add `income_payroll_taxes` to the output dict, using the already-computed `income_tax_ss`.
- **Bug fix:** `income_tax_ss` was computed via `tax.income_tax_liab(...)` but passed `bqssmat` (bequests) as the `n` (labor supply) argument. The TPI equivalent (`income_tax_mat`) correctly passes labor supply. `income_tax_ss` was previously never read, so this had no effect until now — but it needs to be correct before it is surfaced in the output. Fixed to pass `nssmat`.

**TPI (`ogcore/TPI.py`)**
- Add `theta` and `factor` to the output dict (both already in scope). These are constants, so they match their SS counterparts.

After the change both dictionaries expose the same 72 keys.

**Testing**
- Verified key parity between the two dictionaries: both now expose the same 72 keys.
- `pytest "tests/test_SS.py::test_SS_solver[Baseline]"` and `pytest "tests/test_TPI.py::test_run_TPI[Baseline]"` both pass, exercising the modified output dictionaries at runtime. These tests iterate over the expected (pickled) keys, so the newly added keys don't affect them and existing values are unchanged.
- `ruff format --check .` and `ruff check .` clean.

Happy to adjust naming/placement or add regression coverage for the new keys if you'd like.

### DIFF
--- a/ogcore/SS.py
+++ b/ogcore/SS.py
@@ -982,7 +982,7 @@ def SS_solver(
         r_p_ss,
         wss,
         bssmat_s,
-        bqssmat,
+        nssmat,
         factor_ss,
         0,
         None,
@@ -1037,6 +1037,10 @@ def SS_solver(
     yss_before_tax_mat = household.get_y(
         r_p_ss, wss, bssmat_s, nssmat, p, "SS"
     )
+    labor_income_ss = (
+        wss * nssmat * np.squeeze(p.e[-1, :, :]).reshape((p.S, p.J))
+    )
+    capital_income_ss = r_p_ss * bssmat_s
     Css = aggr.get_C(cssmat, p, "SS")
     c_i_ss_mat = household.get_ci(
         cssmat, p_i_ss, p_tilde_ss, p.tau_c[-1, :], p.alpha_c, p.c_min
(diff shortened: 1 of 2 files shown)


jdebacker (approved):

jdebacker:
@arihantlodha-cmd If you can run `make format` to fix code formatting issues, this should be good to go.  Thank you!

arihantlodha-cmd:
Thanks! Synced the branch with `master` and ran `make format` — `ruff format --check .`, `ruff check .`, and `linecheck` all pass now. (The lint failure was the unrelated `tests/BENCHMARK_README.md` on the stale base, fixed on `master` in #79ef1882.) Ready when you are.

jdebacker:
Thanks for the work on OG-Core, @arihantlodha-cmd!

PR #1165, opened 2026-07-03
Allow demographics to vary across income groups

jdebacker:
This PR updates OG-Core to allow demographic parameters (mortality, fertility, immigration rates -- and the resulting population distribution) to vary across lifetime income group (J).

With income specific demographics, the $\lambda_j$ are redefined and no longer will represent the shares of each income group in a model period. Rather, the $\lambda_j$ will define the fraction of new "births" into each lifetime income group.

PR #1187, opened 2026-07-28
Divide bequest pools by actual group populations, not birth shares

### LINKED ISSUE #1186: Bequest receipts don't sum to bequests left when demographics vary by income group
With the new income-group demographics (PR #1165), a model calibrated with a mortality gradient fails the steady-state resource-constraint check. In an OG-ZAF calibration with South African mortality gradients, the violation is about 1% of GDP: `RuntimeError: Steady state aggregate resource constraint not satisfied`, resource constraint ≈ −0.0105.

The cause is in `household.get_bq`, in the `use_zeta = False` branch. Per-capita bequests are computed as `BQ[j] / lambdas[j]`, which treats each group's population share as its birth share. That was exact before #1165. With a mortality gradient, groups no longer have λ-proportional populations — in our calibration the poorest group's actual population share is 0.226 against λ = 0.25, and the top group's is 0.011 against λ = 0.010 — so each group's receipts total `BQ[j] × (actual share / λ_j)` instead of `BQ[j]`. Bequests received no longer equal bequests left: with savings concentrated in the (longer-lived) top groups, households collectively receive about 6.5% more bequests than exist, and the phantom resources show up in the resource constraint.

Reproducer against ogcore 0.18.0 (any Specifications with a mortality gradient):

```python
omega_SS = np.asarray(p.omega_SS)          # (S, J), non-separable under gradients
b = np.outer(np.linspace(0.5, 2.0, p.S),   # savings tilted toward top groups
             np.array([0.2, 0.4, 0.7, 1.2, 2.0, 4.0, 10.0]))
BQ = np.asarray(aggregates.get_BQ(0.04, b, None, p, "SS", False)).ravel()
received = sum((household.get_bq(BQ, j, p, "SS") * omega_SS[:, j]).sum()
               for j in range(p.J))
# BQ.sum() = 0.031880, received = 0.033957  ->  +6.5% over-distribution
```

The `use_zeta = True` branch was already updated in #1165 — it divides by the actual `omega_SS[:, j]` / `omega` — so the fix is to give the non-zeta branch the same treatment: divide each group's pool by its actual population share (`omega_SS[:, j].sum()` in the SS, and the per-period group shares along the time path) instead of `lambdas[j]`. With demographics common across groups the two coincide, so existing results are unchanged.

draft PR to follow.  cc: @rickecon @jdebacker

marcelolafleur:
Fixes #1186.

With demographics that vary across lifetime income groups (#1165), each group's population share no longer equals its birth share `lambdas[j]` and survivorship differs by group. The `use_zeta = False` branch of `household.get_bq` still divided each group's bequest pool by `lambdas[j]`, so bequests received did not sum to bequests left. In an OG-ZAF calibration with South African mortality gradients, households collectively received about 6.5% more bequests than existed, and the steady state failed the aggregate resource constraint by about 1% of GDP.

**The fix**: divide each group's pool by its actual population share: `omega_SS[:, j].sum()` in the steady state, per-period `omega` sums along the time path,  which is what the `use_zeta = True` branch already does. When demographics are common across groups the two denominators coincide, so results are unchanged in every pre-#1165 use case: all existing `get_bq` tests pass without modification.

**Tests**: adds a conservation test with a non-separable `omega` (group population shares deliberately different from `lambdas`) checking that receipts aggregate back to the pool for SS and TPI, by-group and pooled. `tests/test_household.py` passes 52/52 locally.

cc: @rickecon @jdebacker

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Bug Fixes
+
+- Fixes Issue [#1186](https://github.com/PSLmodels/OG-Core/issues/1186):
+  with demographics that vary across lifetime income groups (PR #1165), the
+  `use_zeta = False` branch of `household.get_bq` divided each group's
+  bequest pool by its birth share (`lambdas[j]`) rather than its actual
+  population share, so bequests received did not sum to bequests left and
+  the steady-state resource constraint failed. Receipts are now divided by
+  the group's actual population (from `omega_SS` / `omega`), matching the
+  `use_zeta = True` branch. Results are unchanged when demographics are
+  common across groups.
+
 ## [0.18.1] - 2026-07-22 12:00:00
 
 ### Bug Fixes
--- a/ogcore/household.py
+++ b/ogcore/household.py
@@ -192,21 +192,30 @@ def get_bq(BQ, j, p, method):
                     * utils.to_timepath_shape(BQ)
                 ) / p.omega[:len_T, :, :]
     else:
+        # Divide each group's bequest pool by its actual population share
+        # rather than its birth share (lambdas): with demographic gradients
+        # across income groups (PR #1165), survivorship differs by group and
+        # the two are no longer equal. With common demographics they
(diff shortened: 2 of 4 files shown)


jdebacker (approved):

jdebacker:
@copilot resolve the merge conflicts in this pull request

jdebacker:
@marcelolafleur Thanks for catching this!  Once conflicts are resolved, I will merge.

marcelolafleur:
@jdebacker thanks for the review. Conflicts are fixed, just waiting for the tests to pass

PR #1190, opened 2026-07-29
Make NDC pension growth-rate settings parameters so the system can run

### LINKED ISSUE #1169: NDC pension system cannot run: growth-rate settings are not parameters
Found while fixing #1075 / working #1014 (PR #1167): the NDC functions read `p.ndc_growth_rate` and `p.dir_growth_rate`, but neither exists in `default_parameters.json`, so `pension_system = "Notional Defined Contribution"` fails on any real Specifications object with an AttributeError before reaching the numba-loop problems fixed in that PR. Making NDC usable needs those settings added as parameters (with choices "r", "Curr GDP", "LR GDP") plus the same real-object tests the DB and PS systems now have.

marcelolafleur:
Fixes #1169. Follow-up to #1167, which fixed the other two non-US pension systems.

The notional defined contribution pension system crashes as soon as you try to use it. Its growth-rate settings were never added to the parameters file, so the model stops with an error before computing anything. Behind that error sat two smaller bugs of the same kind: the growth-rate functions expect time paths but get single numbers in the steady state, and the benefit formula reads mortality rates from an attribute that only the unit tests ever set.

Changes:

- The growth-rate settings are now real parameters. The options are the interest rate, current GDP growth, or long-run GDP growth. The default is long-run GDP growth, which is what the code already fell back to, so existing runs are unchanged.
- The points system's growth-rate setting is added as a parameter too, since it has the same problem waiting.
- The growth-rate functions now work in both the steady state and on the transition path.
- The benefit formula uses the model's own mortality rates when the test-only attribute is missing.
- The NDC system is added to the test that runs each pension system on a real parameters object. The contribution rate must be set above its default of zero, or the pension is zero everywhere.

Tested: pension and parameter test suites pass, ruff clean. Bumps to 0.19.0 with a changelog entry.

cc @rickecon @jdebacker

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,22 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.19.0] - 2026-07-29 12:00:00
+
+### Added
+
+- Fixes Issue [#1169](https://github.com/PSLmodels/OG-Core/issues/1169):
+  the Notional Defined Contribution pension system could not run against
+  a real `Specifications` object because its growth-rate settings were
+  not parameters. `ndc_growth_rate`, `dir_growth_rate`, and
+  `points_growth_rate` are now parameters in `default_parameters.json`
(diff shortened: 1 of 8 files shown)


jdebacker (approved):

jdebacker:
@marcelolafleur Thank you for the fix for these pension systems.

Issue #985, opened 2024-09-09
Relax assumption on government consumption good production


jdebacker:
We currently assume that goods/services making up government consumption expenditures are produced from just one industry (namely, industry $M$).  

However, given data from an input/output table, it would be trivial to find demand for output from any sector $m$ that results from gov't consumption expenditures. 

Analogous to how we model differentiated consumption goods, we can have a unit of gov't expenditures be made up of outputs from different sectors using a fixed-coefficient matrix.  E.g., demand for industry $m$ output from gov't consumption will be given as:

$$
G_{m,t} = \pi_{gov, m} G_{t}
$$

where $\pi_{gov, m}$ is the fraction of a unit of gov't spending that comes from industry $m$.

The price of a unit of gov't consumption will then be:

$$
p_{gov,t} = \sum_{m=1}^M   \pi_{gov, m} p_{m}
$$

Government spending in the government's budget constraint, currently $G_{t}$ will be replaced by $p_{gov,t}G_{t}$ since this change will mean that gov't expenditures are not only on the numeraire good (which is the current assumption in OG-Core).

PR #1192, opened 2026-08-03
Allows government consumption and investment to come from multiple sectors

jdebacker:
In the current OG-Core, the M-th industry produces all of the non-consumption goods: capital, government consumption, and infrastructure.  This PR relaxes the assumption somewhat, allowing all industries' output to combine in producing government consumption expenditures and infrastructure goods.  

Addresses Issue #985


jdebacker:
One important note on the changes in this PR: because each industry contributes to the production of `G` and `I_g` we need to know these quantities when solving the firm's problem. But `G` (under the unbalanced budget parameterization and once the closure rule kicks in) can't be determined without `total_tax_revenue`, which itself depends on the solution to the firm's problem. Therefore, the changes here modify the solution algorithm by adding `G` to the "outer loop" variables to be solved for in the GE fixed point algorithm.

This also means that `G` is now passed to some functions where it wasn't previously (such as `SS.inner_loop`).

When running local tests, I'm finding that all the tests that run fully through the SS solution (except for one case) pass:
```
====================================== short test summary info ======================================
FAILED tests/test_SS.py::test_run_SS[Reform, small open use zeta] - RuntimeError: Steady state equilibrium not found
====================== 1 failed, 14 passed, 24 warnings in 16377.51s (4:32:57) ======================
```

and that all the tests that run fully through the TPI solution (except for the new test added where the IO matrix represents multiple industries contributing to `G` and `I_g`) pass:
```
FAILED tests/test_TPI.py::test_run_TPI_full_run[Baseline, mixed government IO] - RuntimeError: Steady state aggregate resource constraint not satisfied
============ 1 failed, 12 passed, 11 warnings in 7226.51s (2:00:26) ===========
```

I'm going to ensure these tests pass. But then for other tests that add a new in put (usually G, such as in `SS.inner_loop`) I will feel confident changing the expected output to match the new output since we have run the changes through the full SS and transition path solutions and found that they do not affect the results.  I will document the cases/tests where expected output changed in this thread.

jdebacker:
Just one local test failing right now:
```
FAILED tests/test_SS.py::test_run_SS[Reform, small open use zeta] - RuntimeError: Steady state equilibrium not found
=========== 1 failed, 643 passed, 77 warnings in 34569.87s (9:36:09) ===========
```

arihantlodha-cmd (commented):
Focused on the new parameters and their docs and validation here, not the full production-sector derivation, so treat this as a partial review.

The io_matrix changes look clean and the docs match the code. I traced the row indexing: the description says the first I rows are consumption goods, row I is government consumption, and row I+1 is infrastructure, and that lines up exactly with how it gets used (io_matrix[:p.I] for C, io_matrix[p.I] for G, io_matrix[p.I+1] for I_g in both SS.py and aggregates). Nice that the new shape validation in parameters.py, checking (I+2, M) and that each row sums to 1, makes a bad matrix fail loudly instead of flowing through silently.

Two small things:

The row-sum check is good, but it doesn't verify entries are non-negative. A hand-built io_matrix could have a negative coefficient and still sum to 1 per row, which would be economically off but pass validation. Worth a >= 0 check while you're already validating shape and sums?

Minor: when the row-sum check fails the error just says each row must sum to 1. Naming the offending row index would make a mis-entered matrix faster to fix, but not important.

Also looked at the new p_g and p_Ig entries in model_variables.json. Labels are escape-clean and the trailing-newline fix on that file is a nice touch.

Looks good from the parameter side.

rickecon:
@jdebacker. Can you update this PR with the most recent version of the OG-Core `master` branch?

vahid-ahmadi:
Large PR, so I focused on the piece most likely to bite silently, which isn't in the diff itself.

`fiscal.py` still sets `G = p.alpha_G[:T] * Y[:T]` and `I_g = p.alpha_I[...] * Y`, i.e. quantities, while the budget constraint now charges `p_g * G` and `p_Ig * I_g`. Because `p_m` is normalized to industry M, `p_g = 1` exactly under the old all-from-M assumption but not once the government row mixes industries. So `alpha_G` and `alpha_I` quietly change meaning from spending-to-GDP to quantity-to-GDP, and any calibration that set them from observed government spending shares now delivers `p_g * alpha_G` of GDP instead. Worth either renormalizing inside `get_G_ss`/`D_G_path` or flagging it prominently in the release notes — it's the kind of thing that shifts every downstream country model's fiscal baseline without failing anything.

Separately, the `(I+2, M)` shape check in `compute_default_params` will raise on every downstream calibration passing today's `I x M` matrix. Appending two copies of `e_M` when an `I x M` matrix arrives would reproduce the old behaviour exactly and keep OG-USA/OG-PHL running through the transition.

Defaults do preserve `p_g = 1`, and `I_d_vec` / `net_capital_outflows_vec` are still tied to industry M, which matches the stated scope.

PR #1195, opened 2026-08-09
Add npv_table for net present value of flow-variable changes

### LINKED ISSUE #1131: Include calculations of the net present value of flows in `output_tables.py`
It's proved useful in many projects to compute the next present value of the changes in flow variables such as GDP. It'd be helpful to easily create such a table with `output_tables.py`.

The cost could look like something from the [Cost of Disease repo](https://github.com/OpenSourceEcon/CostOfDisease/blob/main/code/create_plots_tables.py):
```python
# Find NPV of levels of GDP over NUM_YEARS_NPV years
    results_NPV = {"Discount Rate": [r"1\%", r"2\%", r"3\%", r"4\%", r"6\%"]}
    npv_dict = {
        "Discount Rate": [0.01, 0.02, 0.03, 0.04, 0.06],
        "Discount Rate Label": [r"1\%", r"2\%", r"3\%", r"4\%", r"6\%"],
    }
    for k in reform_dict.keys():
        results_NPV[k] = []
        for r in npv_dict["Discount Rate"]:
            results_NPV[k].append(
                (
                    GDP_series["Diffs"][k][:NUM_YEARS_NPV]
                    / (1 + r) ** np.arange(NUM_YEARS_NPV)
                ).sum()
                / 1e9  # convert to billions of dollars
            )
```

arihantlodha-cmd:
Closes #1131.

Adds `output_tables.npv_table`, which builds a table of the net present value of the reform-minus-baseline change in flow variables (e.g. `Y`) over a horizon, evaluated at a list of discount rates. For each variable:

```
NPV = sum_{t=0}^{num_years-1} (x_reform[t] - x_base[t]) / (1 + r)^t
```

This follows the pattern the issue points to from the Cost of Disease repo, generalized into a reusable table function that fits the existing `output_tables.py` conventions (same `base_tpi`/`reform_tpi`/`*_params` signature, `VAR_LABELS`, and `save_return_table` for csv/tex/excel/json output).

Design choices, flagged for review:

- **Un-stationarized by default.** Discounting the stationarized (detrended) values isn't the economically meaningful object, so the variables are un-stationarized before discounting so the NPV runs over the actual trend-inclusive level path. `stationarized=True` keeps the model values if a caller wants them. (This is why the default differs from `macro_table`, which defaults to stationarized.)
- **Reform required.** The table is defined on a reform-minus-baseline change, so `reform_tpi` is asserted to be present.
- **Model units.** Results are in the variable's model units. Since #1185 added `factor` to the output dictionaries, a caller can scale by it for dollars; I left that to the caller rather than bake in a dollar-year assumption. Happy to add an optional dollar conversion if you'd prefer it built in.

Output layout: one row per variable, one column per discount rate.

**Testing**
- `test_npv_table` builds the table from the cached run output and checks the shape.
- `test_npv_table_values` checks the discounted sum against a hand calculation at r = 0 and r = 0.1.
- Also verified manually on the cached baseline/reform that NPV falls as the discount rate rises, as expected.
- `pytest tests/test_output_tables.py` passes (21); `ruff format --check .`, `ruff check .`, and `linecheck` clean.


jdebacker (approved):

jdebacker:
@arihantlodha-cmd Thanks for this PR.  This table is helpful. I think we'll want to add some conversion to local currency units, but we can make that in another PR. It's helpful to have this is a starting point.

vahid-ahmadi:
Two paths here return silently wrong numbers rather than erroring.

`start_index = start_year - base_params.start_year` isn't bounds-checked. When `DEFAULT_START_YEAR` is earlier than a country model's `start_year` — routine in OG-PHL and OG-ZAF — the negative index makes the slice wrap or come back empty, and the NPVs are garbage or 0.0. The `num_years <= base_params.T` assert is the wrong bound; you want `0 <= start_index` and `start_index + num_years <= T`.

Second, `utils.unstationarize_vars` falls through to an unchanged `else` for any variable outside its two growth lists, so the un-stationarized default quietly discounts stationarized values for things like `r` or `p_m` with no warning. Since the trend-inclusive path is the whole point of that default, an assert that `v` is in the level-growth list seems worth it. `test_npv_table_values` passes `stationarized=True`, so the branch that actually ships is only shape-tested.

On the plus side: unlike `macro_table`, this doesn't write un-stationarized values back into the caller's `base_tpi`/`reform_tpi` dicts, which is the right call.

arihantlodha-cmd:
Thanks Jason! Good call on the currency conversion. I have a small follow-up ready that adds an optional `factor` argument to `npv_table` for exactly that (reporting in local currency units instead of model units), and I will open it once this one merges so it goes in cleanly on top.

jdebacker (approved):

PR #1196, opened 2026-08-09
Add Japan (392) and Fiji (242) to the offline demographics country map

arihantlodha-cmd:
Adds `"392": "JPN"` to the `country_dict` in `ogcore/demographics.py`, so the offline data fallback (used when no UN API token is present) can find Japan's demographic data by country code, like the other supported countries.

This pairs with **EAPD-DRB/Population-Data#14**, which adds the Japan data (`Data/JPN/`) to the mirror this fallback reads from. That PR should merge first; on its own this entry has no data behind it yet.

Motivation: I'm building an open-source OG model of Japan on top of OG-Core, and Japan is currently the only major economy without an OG country model. This (plus the data PR) lets OG-Core use Japan demographics offline. With a UN API token the live pull already works for Japan (code 392); this just extends the token-free path.

### DIFF
--- a/ogcore/demographics.py
+++ b/ogcore/demographics.py
@@ -130,6 +130,7 @@ def get_un_data(
             "076": "BRA",
             "410": "KOR",
             "231": "ETH",
+            "392": "JPN",
         }
         un_variable_dict = {
             "68": "fertility_rates",
--- a/ogcore/demographics.py
+++ b/ogcore/demographics.py
@@ -131,6 +131,7 @@ def get_un_data(
             "410": "KOR",
             "231": "ETH",
             "392": "JPN",
+            "242": "FJI",
         }
         un_variable_dict = {
             "68": "fertility_rates",


marcelolafleur:
Thanks @arihantlodha-cmd — the entry looks right. One note for whoever merges: `Data/JPN` doesn't exist on the Population-Data side yet, so this would 404 today. Being sorted in EAPD-DRB/Population-Data#14.

Also, while you're in this dict — worth adding `"242": "FJI"`. Fiji landed in the archive this morning and this map missed it, so the offline path KeyErrors on Fiji right now.

arihantlodha-cmd:
thanks @marcelolafleur. good news on the dependency: Population-Data#14 is merged and the generator has run, so `Data/JPN` now exists in the mirror with the right 2020-2099 range. this entry should resolve cleanly now.

and done on Fiji: added `"242": "FJI"` in the same commit. confirmed FJI is in the archive but was missing here, so the offline path was KeyErroring on it. both Japan and Fiji now have data behind their entries.

marcelolafleur:
@arihantlodha-cmd Japan and Fiji are both added and I also fixed all the scripts to keep that repo up to date. This is ready to close @jdebacker @rickecon

rickecon (approved):

rickecon:
@arihantlodha-cmd. Thanks for this PR. Looks good. Merging now.

PR #1201, opened 2026-08-11
Apply replacement_rate_adjust to every pension system

### LINKED ISSUE #1200: replacement_rate_adjust is ignored by every pension system except US-Style Social Security
`replacement_rate_adjust` is a `(T+S, J)` multiplier on the pension replacement rate — the natural way to model a country that automatically cuts benefits as it ages.

It only works under one of the four pension systems. Under the other three it is accepted, stored, and ignored.

**The test.** Same parameter, same call, two systems:

```python
out = pensions.pension_amount(0.04, 1.0, n, 1.0, 0.1, -1, 0, False, "SS", e, 1e5, p)
```

```
US-Style Social Security   adj=1.0 -> 3.60000   adj=0.5 -> 1.80000   scales exactly
Defined Benefits           adj=1.0 -> 0.92659   adj=0.5 -> 0.92659   no effect
```

**Why.** In `ogcore/pensions.py` the parameter is read in six places, all inside `SS_amount()` (lines 159–199), which serves only `"US-Style Social Security"`. `DB_amount()` (line 204), `NDC_amount()` (332) and `PS_amount()` (429) contain no reference to it.

**Why it matters.** Nothing warns. A calibration that sets it under Defined Benefits looks calibrated and is not — the solved steady state is byte-identical to one that never set it (`pension/Y = 0.093040`, `theta = 0.0038901`, both ways). This is the same trap as `alpha_db` defaulting to `0.0`, where switching to Defined Benefits without setting it pays zero pensions. A parameter that silently does nothing is worse than a missing one, because it produces a confident wrong answer.

**A concrete case.** Japan's pension system has a *macroeconomic slide* that automatically reduces benefits to stay solvent — the MHLW 2024 actuarial review projects the replacement rate falling from 61.2% in FY2024 to 50.4% in FY2057. Without a working adjustment, the DB block pays a fixed replacement rate, so the pension bill rises mechanically with the old-age ratio: 12.6% of GDP against an actual 9.3% today. The calibration then has to choose between matching today's spending and matching the long run, when Japan's actual institution does both. Any country with a legislated benefit glide or a notional-account balancing mechanism hits this.

**Suggested fix.** Move the adjustment out of `SS_amount` and into `pension_amount`, which already receives `t` and `method`. Every system then honours it, and US-Style results are unchanged because the steady state keeps taking the terminal value:

```python
def pension_amount(r, w, n, Y, theta, t, j, shift, method, e, factor, p):
    if p.pension_system == "US-Style Social Security":
        pension = SS_amount(...)      # with its private copy removed
    elif ...

    adj = p.replacement_rate_adjust
    idx = -1 if method == "SS" else t
    scale = adj[idx, :] if j is None else adj[idx, j]
    return pension * scale
```

happy to open a PR.

cc @jdebacker @rickecon

marcelolafleur:
Fixes #1200.

`replacement_rate_adjust` is a table of multipliers on pension benefits, one per year and income group. It lets a country phase benefits down over time.

It was only read inside `SS_amount`, so it worked under US-Style Social Security and did nothing at all under Defined Benefits, Notional Defined Contribution and Points System. Setting it under those three changed nothing and gave no warning.

`pension_amount` now applies it to the other three systems as well.

The only fiddly part is looking up the right year. In the steady state there is one year, so the multiplier is the last row of the table. Along the transition the model computes a block of years at once, and each row of that block is a different year — the block starts at year `t`, so its first row needs the multiplier for year `t`, its second row year `t + 1`, and so on. A small helper does that lookup and returns the multipliers shaped to line up with the benefits array. This is what `SS_amount` was already doing for US-Style; the helper does the same for the rest.

`SS_amount` itself is unchanged, so US-Style results cannot move.

With this fix, thame call, with the multiplier halved:

```
                     mult=1.0    mult=0.5
US-Style              3.60000     1.80000
Defined Benefits      0.92659     0.46329     (before this was 0.92659 as well)
```

### Tests

- `test_replacement_rate_adjust_applies_to_every_system` : halving the multiplier must halve benefits, under all four systems. Fails on master for DB, NDC and PS.
- `test_replacement_rate_adjustment_indexes_by_cohort_year` : checks the year lookup: on the transition, row *n* of the block must get the multiplier for year `t + n`, and the steady state must get the last row. Using one year's multiplier for the whole block would still pass a test with a flat table, and would quietly cancel out a phase-down. The existing `test_pension_amount` sets the table to all ones, so it cannot catch this.

`test_pensions.py`, `test_tax.py` and `test_aggregates.py` pass, 122 total.

cc: @jdebacker @rickecon


arihantlodha-cmd (commented):
Nice, this is a clean fix for a genuinely sneaky bug. A parameter that silently does nothing under three of the four systems, with no warning, is exactly the kind of thing that bites someone quietly.

I'm still newer here so treat this as non-blocking, but I traced the indexing since that's the load-bearing claim. The helper matches `SS_amount` row for row: SS uses `adjust[-1]`, the 1-D time-path case uses `adjust[t]`, and the 2-D case's `adjust[t:t+length]` is exactly the `t + tt` per-cohort rows `SS_amount` loops over at line 70. So the "mirrors SS_amount" claim holds. And nice that you tested both halves: one test that the adjustment actually applies under every system, and a separate one that pins the per-cohort offset directly, which is the part most likely to hide an off-by-one.

One thing worth making explicit. The helper multiplies the whole `pension` array, whereas `SS_amount` only touches `pension[retireTPI:]`. Those are equivalent only because DB/NDC/PS return zero benefits before retirement, so scaling the zeros is harmless. That holds today, but it's an implicit assumption a future pension formula could break. A one-line comment, or an assert in the test that the pre-retirement entries stay zero, would keep that from quietly going wrong.

Minor: `TPI_scalar` returns `adjust[0, j]`, the first year. Is row 0 the intended year for the scalar case, or should it track the period being computed like the other branches?

Thanks for catching this one.

marcelolafleur:
Thanks @arihantlodha-cmd, good catch on the implicit assumption. I added a note to the helper's docstring: scaling the whole array only works because DB, NDC, and PS pay nothing before retirement, and a future system that pays earlier would need the SS_amount treatment.

On TPI_scalar: row 0 is intended. That branch exists for exactly one caller, firstdoughnutring, which handles one special household: the oldest person alive when the transition begins, who has just one period left to live. Their entire remaining problem happens in year 0, by construction, so row 0 is always the right row. It also matches what SS_amount already does for this case.

arihantlodha-cmd:
thanks @marcelolafleur, the docstring note is exactly the right guard. and the TPI_scalar explanation clicks: if firstdoughnutring's one household has a single period left and lives it entirely in year 0, then row 0 is unambiguous by construction. looks good to me.

vahid-ahmadi:
The helper's `ndim == 2` branch slices `adjust[t : t + length, j]`, treating axis 0 as time — but `DB_amount`/`NDC_amount`/`PS_amount` return `(S, J)` age-major for 2-D `n`. No current caller reaches it (TPI calls are 1-D per-cohort or 3-D aggregate), so it's latent rather than live, but it's worth an assert or a comment before someone wires up a 2-D path.

Related gap: `NDC_amount` and `PS_amount` have no `ndim == 3` branch at all, so the helper's 3-D path is only live for DB and the `aggregates.py` TPI call would already fail for those two systems. The tests don't currently reach it either — the equivalence test runs only `method="SS"` through `pension_amount`, and the indexing test calls the helper directly on synthetic arrays, so there's no TPI path through `pension_amount` covered for any of the three systems this fixes.

jdebacker (approved):

PR #1209, opened 2026-08-26
Allow a supplied initial distribution of wealth in the transition

arihantlodha-cmd:
The time path solver initializes the t=0 distribution of household wealth from the steady-state distribution, scaled so its aggregate matches the steady-state B. That assumes the economy starts already at its steady-state age-wealth profile. For a country whose model start-year age structure is far from its steady state, such as a rapidly aging population, that profile is a poor initial condition and can leave a small resource-constraint residual in the first periods of the path.

This adds an optional hook to supply an observed initial distribution instead. A new boolean `use_initial_b_dist` gates a new SxJ array `initial_b_dist`; when the flag is set, `get_initial_SS_values` takes the shape of the initial wealth distribution from `initial_b_dist` rather than the steady state. The aggregate is still rescaled to the steady-state B, so only the cross-sectional shape is overridden. The default leaves the flag off and reproduces the previous `initial_b` exactly, so existing runs are unchanged.

Adds `test_get_initial_SS_values_custom_dist` covering the aggregate-preserving and shape-override behavior; the existing `test_get_initial_SS_values` regression cases and the parameter/util tests pass unchanged.

### DIFF
--- a/ogcore/TPI.py
+++ b/ogcore/TPI.py
@@ -161,8 +161,20 @@ def get_initial_SS_values(p):
     baseline_ss = os.path.join(p.baseline_dir, "SS", "SS_vars.pkl")
     ss_baseline_vars = utils.safe_read_pickle(baseline_ss)
     factor = ss_baseline_vars["factor"]
-    B0 = aggr.get_B(ss_baseline_vars["b_sp1"], p, "SS", True)
-    initial_b = ss_baseline_vars["b_sp1"] * (ss_baseline_vars["B"] / B0)
+    # Shape of the initial (t=0) distribution of wealth across ages and
+    # ability types. By default this is the steady-state distribution
+    # (ss_baseline_vars["b_sp1"]), which assumes the economy begins already
+    # at its steady-state age-wealth profile. A country model whose starting
+    # age structure is far from its steady state (e.g., a rapidly aging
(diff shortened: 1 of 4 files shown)


vahid-ahmadi:
The default `initial_b_dist` is `[[0.0]]`, which `extrapolate_array` expands to an all-zero `(S, J)`. So setting `use_initial_b_dist=True` without also supplying the array gives `B0 = 0` and then `b_dist * (B / 0)` — silent NaNs propagating into `b_sinit`, `BQ0` and the whole path rather than an error. A validator, or an assert that the supplied distribution is nonnegative with a positive aggregate, would catch that at the point the user made the mistake.

Worth being explicit about what this does and doesn't change: `B0 = get_B(b_dist, ...)` and `initial_b = b_dist * (B / B0)`, so by linearity `B_init[0]` equals `ss_vars["B"]` exactly either way, and `K_init[0] = B_init[0] * K/B` with it. Only the cross-sectional shape moves. That's a reasonable thing to want, but it isn't the aggregate-level channel the description motivates — might be worth restating so users don't expect a level effect.

Minor: `extrapolate_array` with 2-D dims fills forward along axis 0 as a time path, so supplying fewer than S rows silently repeats the last age rather than erroring.

Issue #1211, opened 2026-08-28
Dask client handling in TPI/SS: unguarded scatter breaks serial runs; SS re-scatters params every residual evaluation


vahid-ahmadi:
## 1. `run_TPI` raises when `client=None`

`client.scatter` at [`TPI.py:969`](https://github.com/PSLmodels/OG-Core/blob/master/ogcore/TPI.py#L969) is called unconditionally at function-body indentation:

```python
    # Scatter the parameters
    scattered_p_future = client.scatter(p, broadcast=True)
```

`SS.inner_loop` guards the equivalent block with `if client:` ([`SS.py:292`](https://github.com/PSLmodels/OG-Core/blob/master/ogcore/SS.py#L292)), but TPI does not. So `run_TPI(p, client=None)` fails with `AttributeError: 'NoneType' object has no attribute 'scatter'` before the loop is entered — which makes the serial fallback at `TPI.py:1039` unreachable in exactly the case it exists for.

Fix is to wrap the scatter/restore block in `if client:` and pass `p` directly when there is no client, mirroring SS.

## 2. `SS.inner_loop` re-scatters the parameters on every residual evaluation

`client.scatter(p, broadcast=True)` sits **inside** `inner_loop` ([`SS.py:304`](https://github.com/PSLmodels/OG-Core/blob/master/ogcore/SS.py#L304)), which is called once per outer residual evaluation by `SS_fsolve`. The `Specifications` object is therefore re-serialised and re-broadcast to every worker dozens of times per steady-state solve, while `p` does not change across those calls.

`run_TPI` already does this correctly — scatter once before the loop, reuse the future. Hoisting the SS scatter into `run_SS` (or caching the future on the client) would match that pattern and remove the repeated broadcast.

I can open a PR for (1) straight away; happy to do (2) in the same PR or separately, whichever you prefer.

PR #1212, opened 2026-08-28
Guard the Dask scatter in run_TPI so serial (client=None) runs work

vahid-ahmadi:
## Why

`run_TPI(p, client=None)` crashes. `client.scatter` is called unconditionally, so passing no Dask client raises `AttributeError` before the TPI loop is entered — which makes the serial fallback at `TPI.py:1039` unreachable in exactly the case it exists for. `SS.inner_loop` already guards the equivalent block with `if client:`; TPI doesn't.

## What changes after merging

Serial TPI runs work. Anyone calling `run_TPI` without a Dask cluster — scripts, notebooks, CI, small debugging runs — goes from a hard crash to a working solve. No behaviour changes when a client is supplied.

## Change

Wraps the schema-backup / scatter / restore block in `if client:`. `scattered_p_future` is only read inside the `if client:` submit branch, so this is a guard plus an indent — no logic moves.

## Evidence

New `tests/test_TPI.py::test_run_TPI_serial_no_client` seeds SS from the cached `test_io_data` pickles and monkeypatches `inner_loop` to raise a sentinel, asserting execution reaches the first serial household solve. No solve is run; ~1s, no Dask, so it is unmarked and guards this path in normal CI.

```
with the fix:      1 passed in 1.15s
fix stashed:       FAILED - AttributeError: 'NoneType' object has no attribute 'scatter'
```

Fixes the first half of #1211. The second half (SS re-scattering per residual evaluation) is #1214.

### DIFF
--- a/ogcore/TPI.py
+++ b/ogcore/TPI.py
@@ -955,25 +955,26 @@ def run_TPI(p, client=None):
     trust_radius_max = getattr(p, "TPI_trust_radius_max", 10.0)
     prev_accel_dist = np.inf
 
-    # Before scattering, temporarily remove unpicklable schema objects
-    schema_backup = {}
-    for attr in ["_defaults_schema", "_validator_schema", "sel"]:
-        if hasattr(p, attr):
-            schema_backup[attr] = getattr(p, attr)
+    if client:
+        # Before scattering, temporarily remove unpicklable schema objects
+        schema_backup = {}
+        for attr in ["_defaults_schema", "_validator_schema", "sel"]:
+            if hasattr(p, attr):
+                schema_backup[attr] = getattr(p, attr)
+                try:
+                    delattr(p, attr)
+                except Exception:
+                    pass
+
+        # Scatter the parameters
+        scattered_p_future = client.scatter(p, broadcast=True)
+
+        # Restore the schema objects (they're not needed by workers anyway)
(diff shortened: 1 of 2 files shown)


vahid-ahmadi:
On the Codecov flag (8.33% patch coverage, 11 lines missing) — worth explaining, since it reads worse than it is.

The diff is a guard plus a re-indent, so Codecov counts the 11 re-indented lines of the schema-backup/scatter/restore block as new. Those lines only execute **with** a Dask client, and the non-local CI suite runs without one — so they show as uncovered even though they are byte-identical to what was already there and already ran in the `local`-marked tests.

The same report shows `ogcore/TPI.py` coverage going **up 17.17%** (35.25% → 52.42%), which is the real effect: before this change `run_TPI(p, client=None)` raised `AttributeError` before the loop, so the entire serial path was unreachable and uncovered. It now executes.

I have separately verified the serial path is not just reachable but numerically correct: with `ENFORCE_SOLUTION_CHECKS = False`, serial output compared against a real `LocalCluster` run across all 72 keys of the returned dict gives 0 mismatches at 1e-8. The regression test in this PR asserts reachability only, which is the narrow thing that was broken — I did not want to add a multi-minute solve to CI for it.

jdebacker (approved):

Issue #1210, opened 2026-08-28
Resource-constraint check cannot exempt the terminal period without waiving the whole path


vahid-ahmadi:
## Summary

`RC_TPI` is a single scalar compared against the resource-constraint error at **every** period with `np.any`, so there is no way to exempt the terminal period without simultaneously waiving the check across the whole path.

https://github.com/PSLmodels/OG-Core/blob/master/ogcore/TPI.py#L1839

```python
if (np.any(np.absolute(RC_error) >= p.RC_TPI)) and ENFORCE_SOLUTION_CHECKS:
    raise RuntimeError("Transition path equlibrium not found " + "(RC_error)")
```

## Why this matters

The last period of a truncated transition path is not a true steady state, so `RC_error[T-1]` is routinely larger than the interior periods by orders of magnitude. Country packages that hit this have no expressive option other than raising the global tolerance.

A concrete example downstream: OG-UK sets `p.RC_TPI = 0.2` with an in-code comment stating that only `t = T-1` fails and "all other periods are well within 1e-4" (`oguk/api.py:979-983`). The effect is that a mid-path violation of up to 0.2 — the same order as the terminal error the setting was written to route around — now passes silently on every transition run.

## Suggested change

Check the interior and terminal periods separately, e.g.

- compare `RC_error[:-1]` against a tight tolerance (`RC_TPI`, default unchanged), and
- handle `RC_error[-1]` under a separate, looser `RC_TPI_terminal` parameter (or exclude it and note the truncation in the docs).

This is backwards compatible for anyone currently passing the check, and it restores the diagnostic for the interior path.

Happy to open a PR if the approach looks right.

arihantlodha-cmd:
A data point from a second country package: I hit this exact terminal-period behavior building OG-Japan. Across the transition the resource constraint held to ~1e-7 at every interior period, but the final period carried a much larger residual (up to ~5% under a large pension-benefit reform), independent of `TPI_outer_method`. Same signature as the OG-UK numbers here and in #1216, which supports reading it as a truncation/boundary property rather than a solver artifact. I ended up disabling `ENFORCE_SOLUTION_CHECKS` for those runs, which is exactly the "no expressive option" problem described here.

One piece is already in place: #1213 changed the failure message at L1839 to report the period and magnitude of the worst violation, so a terminal-only violation is now distinguishable from an interior one at a glance. That is the diagnostic half; this issue is the enforcement half.

[...]

PR #1213, opened 2026-08-29
Report the size and location of a resource-constraint failure in TPI

arihantlodha-cmd:
When a transition path fails the resource-constraint check, the error is a bare `Transition path equlibrium not found (RC_error)` with no indication of how large the violation is or which period it is in. Working out whether a failure is a benign boundary artifact or a real calibration problem meant reloading the saved `TPI_vars.pkl` and profiling `resource_constraint_error` by hand.

This puts the diagnostic in the message: the maximum absolute resource-constraint error, the period it occurs in, and the tolerance, plus a one-line note on how to read it (a violation confined to the first or last periods is usually an initial- or terminal-boundary artifact; one spread across the path points to an inconsistent calibration). It mirrors the existing outer-loop stall messages, which already name the likely cause. Also fixes the `equlibrium` typo in that message.

Message-only change: it does not alter when the check fires, only what it reports when it does.

### DIFF
--- a/ogcore/TPI.py
+++ b/ogcore/TPI.py
@@ -731,6 +731,36 @@ def inner_loop(guesses, outer_loop_vars, initial_values, ubi, j, ind, p):
     return euler_errors, b_mat, n_mat
 
 
+def _rc_error_message(RC_error, RC_TPI):
+    """
+    Build the message for a transition resource-constraint failure.
+
+    Reports the maximum absolute resource-constraint error, the period it
+    occurs in, and how to read it: a violation confined to the first or last
+    periods is usually an initial- or terminal-boundary artifact, while one
+    spread across the path points to an inconsistent calibration.
+
+    Args:
+        RC_error (array_like): resource constraint error, time on axis 0
+        RC_TPI (scalar): the tolerance the error is checked against
+
+    Returns:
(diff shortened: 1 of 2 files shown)


vahid-ahmadi:
Useful change — I hit exactly this in OG-UK and ended up profiling `resource_constraint_error` out of the pickle by hand, so putting the period and magnitude in the message is worth having.

One thing the reshape does that's worth being deliberate about: `abs_rc.reshape(abs_rc.shape[0], -1).max(axis=1)` collapses everything after the first axis, so `t_worst` is a period index only if axis 0 is time. That holds for the current `RC_error`, but if it ever gains a leading industry or type axis the message would silently report the wrong thing. An assert on `rc_by_period.shape[0] == p.T` would pin it down cheaply.

Related: I filed #1210 on the fact that `RC_TPI` is a single scalar checked with `np.any` across the whole path, so exempting the terminal period means waiving the interior too. This PR makes that much easier to diagnose but doesn't change it — the note about boundary artifacts is essentially describing the limitation. If you're open to it, a separate terminal tolerance would let the interior check stay tight, and this message would then be the thing that tells you which side failed.

jdebacker (approved):

PR #1217, opened 2026-09-02
Add git commands to `Makefile`

jdebacker:
This PR adds helpful git commands to the `Makefile`. 

The changelog is also updated to reflect some recently merged PRs.


rickecon:
@jdebacker. Why do we want this? The `git-sync` command is a shortcut for the standard: `git fetch upstream`, `git merge upstream/master`, `git push origin master`. My initial inclination is to not shortcut the git commands for these operations. My reason is to address the exceptions that come up in a minority of cases.

rickecon:
@jdebacker. It certainly does not harm to have these shortcuts in the Makefile. But I think these are different than other shortcuts in that they circumvent a few Git steps that are actually important for a contributor to know what is going on, even for an experienced constributor.

rickecon:
@jdebacker. If we are keeping these, I would recommend that we rename the files `gitsync.py` and `gitpr.py`.

jdebacker:
@rickecon writes:
>My initial inclination is to not shortcut the git commands for these operations. My reason is to address the exceptions that come up in a minority of cases.

Exceptions would be raised if there is a merge conflict. 

> If we are keeping these, I would recommend that we rename the files gitsync.py and gitpr.py.

Extensions added and files now formatted with ruff.

rickecon (approved):

rickecon:
@jdebacker. This PR looks good. I'll merge it as soon as it passes all the tests.

PR #1220, opened 2026-09-14
Add parameters for calibration of initial distribution of wealth

rickecon:
This PR:

- Adds three new parameters for calibrating the initial distribution of household wealth: `initial_wealth_factor_mat`, `use_initial_BY_ratio`, and `initial_BY_ratio`. The `initial_wealth_factor_mat` adjusts the initial distribution of household wealth relative to the steady-state distribution of household wealth. The other two parameters let the modeler adjust the initial distribution of household wealth by a factor that is based on a target value for the initial aggregate household wealth to GDP ratio. This is an alternative approach to PR #1189 and PR #1209
- Adds documentation in new last section of `calibration.md` on how to calibrate `initial_wealth_factor_mat`, `use_initial_BY_ratio`, and `initial_BY_ratio` parameters.
- Removes the extraneous `requirements.txt` file. This file is clearly out-of-date, and I can't see anything that is calling it or using it. `os` is not a PyPI package, it is Python standard library. `taxcalc` is not a dependency of OG-Core. It omits the important dependencies listed in `pyproject.toml`. `requirements.txt` was added in commit `36727c7c` (Aug 2019) as part of "add readthedocs files," and there's no `.readthedocs.yaml` in the repo anymore — docs now build via the Makefile and workflows. It's been touched twice since, last in `f44266f2` (Jul 2024, "Added numba to requirements.txt"), which looks like someone updating it out of habit rather than because anything consumed it.
- Fixed typos in `output_talbes.py`, `pensions.py`.
- Updates the `uv.lock`.

The following table shows results from different combinations of parameters. All the folders and files are stored in this [`OG_WealthDist_PR` Google Drive folder](https://drive.google.com/drive/folders/1G36znj_7_NwomEYQ0doDIrL41VbD-EQm?usp=sharing).

| Script name | Description | base B_1/Y_1 | base B_SS / Y_SS |
| :--- | :--- | :--- | :--- |
| `run_ogcore_example.py` | Baseline and reform from current master branch | 3.4496 | 3.7954 |
| `run_init_w.py` | Baseline and reform from PR branch with default params | 3.1147 | 3.7954 |
| `run_init_w_usa.py` | Baseline and reform from PR branch with B/Y USA target | ? | 3.7954 |
| `run_init_w_oldmatch.py`| Baseline and reform from PR branch setting B/Y to match old | ? | ? |

cc: @jdebacker @marcelolafleur


jdebacker:
@rickecon Thanks for providing an alternative approach to that in PRs #1189 and #1209.  Calibration of an initial net wealth to GDP ratio will be much easier than calibrating a ratio of initial wealth to steady state wealth.

>Adds three new parameters for calibrating the initial distribution of household wealth: initial_wealth_factor_mat, use_initial_BY_ratio, and initial_BY_ratio. The initial_wealth_factor_mat adjusts the initial distribution of household wealth relative to the steady-state distribution of household wealth. The other two parameters let the modeler adjust the initial distribution of household wealth by a factor that is based on a target value for the initial aggregate household wealth to GDP ratio. 

I would suggests the following modifications here:
1. The `use_initial_BY_ratio` parameter is unnecessary: it would be replaced by `if initial_BY_ratio is None:`.
2. For the same reason that an initial net wealth to GDP makes more sense than a ratio of initial to SS wealth, I think the distributional parameter (`initial_wealth_factor_mat`) should be calibrated without reference to the SS distribution.  I would prose and alternative `initial_wealth_distribution` parameter that is an SxJ matrix showing the fraction of aggregate net worth held by each (s,j) cell in the population.  Such an object could be calibrated in a straight-forward manner from survey data on household net worth (such as the Survey of Consumer Finances in the US) by simply finding the weighted sum of net wealth by age and income group and then dividing that by aggregate net wealth found in the survey ([`wealthinit.py`](https://github.com/PSLmodels/OG-USA/blob/master/ogusa/wealthinit.py) in the OG-USA repo does essentially this).

